PR #4, opened 2024-09-27
[TM-1312] users/me

roguenet:
https://gfw.atlassian.net/browse/TM-1312
[FE PR](https://github.com/wri/wri-terramatch-website/pull/513)


roguenet:
@Scriptmatico Note that the build is failing right now, but I'm going to address that in the next ticket when I work on getting AWS builds up and running. It's not clear to me at a glance why it's not working on this branch as of the last two commits, but linter and tests are passing, which I think is good enough for this ticket.

libs/common/src/lib/decorators/json-api-response.decorator.ts
```diff
@@ -0,0 +1,165 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Scriptmatico, line 1: I understand now why the rule is being disabled, Im wondering if any of those variables can be properly typed. otherwise it's ok.

libs/common/src/lib/util/json-api-builder.ts
```diff
@@ -0,0 +1,142 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Scriptmatico, line 1: Same comment as before if cant be properly typed is ok for now.

libs/common/src/lib/util/tm-log.service.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Scriptmatico, line 1: I think here message can be typed as string or object.

Scriptmatico (approved):

libs/common/src/lib/decorators/json-api-response.decorator.ts
roguenet: I did spend some time looking into that and didn't find a clean way. I avoid disabling this rule whenever possible.

libs/common/src/lib/util/tm-log.service.ts
roguenet: Done. Unfortunately, I had to leave the linter rule disabled because of the optionalParams

PR #25, opened 2024-12-06
[TM-1531] delayed job with data

cesarLima1:
https://gfw.atlassian.net/browse/TM-1531
this PR depends on backend changes in branch feat/TM-1531-delayed-job-with-data


yarn.lock
```diff
@@ -0,0 +1,8973 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@adobe/css-tools@^4.0.1":
```
roguenet, line 5: This repo uses `npm`, not `yarn`. Please don't check in this file.

libs/common/src/lib/dto/user.dto.ts
```diff
@@ -47,4 +47,7 @@ export class UserDto extends JsonApiAttributes<UserDto> {
 
   @ApiProperty({ type: () => UserFramework, isArray: true })
   frameworks: UserFramework[];
+
+  @ApiProperty({ nullable: true })
+  id: number | null;
```
roguenet, line 52: Why is the user id needed by the client?

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -11,26 +12,50 @@ import { DelayedJob } from '@terramatch-microservices/database/entities';
...
+    description: 'Job with that UUID not found.',
   })
   async findOne(@Param('uuid') pathUUID: string): Promise<JsonApiDocument> {
-    const job = await DelayedJob.findOne({ where: { uuid: pathUUID }});
+    const job = await DelayedJob.findOne({ where: { uuid: pathUUID } });
     if (job == null) throw new NotFoundException();
 
-    // Note: Since jobs are very generic and we don't track which resources are related to a given
-    // job, there is no effective way to make a policy for jobs until we expand the service to
-    // include an owner ID on the job table.
```
roguenet, line 32: Please don't remove this comment without implementing a policy to replace it. Now that we have a `createdBy` on delayed jobs, creating a policy is possible and should potentially be done by this ticket. But if it's not done by this ticket, this comment is still needed as a reminder to do it in the future.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -11,26 +12,50 @@ import { DelayedJob } from '@terramatch-microservices/database/entities';
...
+  @JsonApiResponse({ data: { type: DelayedJobDto } })
+  @ApiException(() => UnauthorizedException, {
+    description: 'Authentication failed.',
+  })
+  async getRunningJobs(
+    @Request() { authenticatedUserId }
+  ): Promise<JsonApiDocument> {
+    const runningJobs = await DelayedJob.findAll({
+      where: {
+        isCleared: 0,
```
roguenet, line 29: This should be `isCleared: false`

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -26,13 +26,21 @@ export class DelayedJob extends Model<DelayedJob> {
...
   @Column(STRING)
-  proccess_message: string | null
+  progressMessage: string | null
+
+  @AllowNull
+  @Column(STRING)
+  createdBy: string | null;
+
+  @Column(BOOLEAN)
+  isCleared: boolean;
```
roguenet, line 44: I'm curious about this new column: when will it get set to `false`?

egrojMonroy: There is a button in the frontend inside the notification modal. We need to display all jobs, including completed ones, but once cleared, completed jobs should no longer be shown.

yarn.lock
egrojMonroy: removed

libs/database/src/lib/entities/delayed-job.entity.ts
roguenet: And where does the button call to? All APIs directly related to delayed jobs should be here in the jobs service.

libs/common/src/lib/dto/user.dto.ts
egrojMonroy: Removed, rethinking it, it wont be necesary as the authenticatedId.

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: Rollback erased.

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: fixed

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: Adding the endpoint for the cleared ones.
You are right, this is only intended to display certain ones that comes from the site. 
For this case, what would you advice? Adding a type for delayed job? or name ?

egrojMonroy: The endpoint to clear is done

roguenet: > Adding the endpoint for the cleared ones. You are right, this is only intended to display certain ones that comes from the site. For this case, what would you advice? Adding a type for delayed job? or name ?

Maybe making sure `isClear` defaults to `true` is enough, and then the jobs that _should_ show up here set it to `false` when they're created. It's a bit weird though. Maybe changing the name to `isAcknowledged` would make it a bit more clear. This is a new pattern to me. When I've used systems like this in the past, there was never a need to have a panel that shows the progress of old jobs. For a system like this, the majority of jobs should finish within 30 seconds or a minute, and having a panel to show progress of them is unusual.

egrojMonroy: Great, then in API will change that so the default and change the attrib name.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -11,28 +12,78 @@ import { DelayedJob } from '@terramatch-microservices/database/entities';
...
+
+  @Patch('clear')
+  @ApiOperation({
+    operationId: 'clearNonPendingJobs',
+    description: 'Set isCleared to true for all jobs where status is not pending.',
+  })
+  @ApiException(() => UnauthorizedException, {
+    description: 'Authentication failed.',
+  })
+  async clearNonPendingJobs(@Request() { authenticatedUserId }): Promise<{ message: string }> {
```
roguenet, line 75: I think this blanket "clear all" approach is likely to lead to bugs, and is not very RESTful. It would be much better for the FE to send the list of jobs it would like to clear out. This should be done either as a series of calls to a `@Patch(':uuid')` endpoint, or as a bulk operation. See the research service for  an example of a bulk patch operation. The payload in either case should be sent in JsonAPI format, correctly highlighting the resources by ID and the field(s) that should be updated.

Another possible approach to this problem would be to remove the `isCleared` flag, and do a `Delete()` endpoint instead. One logical problem with this endpoint as written is that it's possible to "clear" a job that hasn't been finished yet, which seems odd to me. Instead, we could do a delete endpoint that takes a set of job UUIDs a query params and deletes them. In that case (and maybe even if we keep this `isCleared` pattern), the endpoint should throw an error if the FE tries to delete / clear a job that is still pending.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -11,28 +12,78 @@ import { DelayedJob } from '@terramatch-microservices/database/entities';
...
+      {
+        where: {
+          isCleared: false,
+          status: { [Op.ne]: 'pending' },
+          createdBy: authenticatedUserId,
+        },
+      }
+    );
+
+    return { message: `${updatedCount[0]} jobs have been cleared.` };
```
roguenet, line 87: No endpoint in v3 should be sending a plain JSON response like this. If a response is required, it must be in JsonAPI format. In the case of a Patch operation, the expected return value would be the modified resource(s), which the FE Connection system will know how to fold into its cache of the server data so that those resources will be updated client side and any component that relies on them will get automatically updated with the refreshed data.

Note that in the case of the research service, I left off the response value. I possibly shouldn't have done that, but that API is different, not being consumed by our FE API layer.

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -26,13 +26,21 @@ export class DelayedJob extends Model<DelayedJob> {
...
+  processedContent: number | null;
 
   @AllowNull
   @Column(STRING)
-  proccess_message: string | null
+  progressMessage: string | null
+
+  @AllowNull
+  @Column(STRING)
+  createdBy: string | null;
```
roguenet, line 41: Why is this column a string? It should be an unsigned bigint foreign key pointing to the users table.

libs/database/src/lib/factories/delayed-job.factory.ts
```diff
@@ -6,7 +6,9 @@ export const DelayedJobFactory = FactoryGirl.define(DelayedJob, async () => ({
...
   statusCode: 200,
   payload: { "data": "test" },
-  total_content: 0,
-  processed_content: 0,
-  proccess_message: "test"
+  totalContent: 0,
+  processedContent: 0,
+  progressMessage: "test",
+  isCleared: false,
+  createdBy: '1290'
```
roguenet, line 13: As noted above, this should be an integer, not a string.

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: I change it to a bulk operation, considering the API format for the input.

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: the response is now the modified resources

libs/database/src/lib/factories/delayed-job.factory.ts
egrojMonroy: Changed for all parts

apps/job-service/src/jobs/dto/delayed-job-update.dto.ts
```diff
@@ -0,0 +1,31 @@
...
+
+class JobAttributes {
+  @IsBoolean()
+  @ApiProperty({ description: 'Value to set for isAcknowledged', example: true })
+  isAcknowledged: boolean;
+}
+
+export class JobData {
+  @ApiProperty({ enum: ['jobs'], description: 'Type of the resource', example: 'jobs' })
+  type: 'jobs';
```
roguenet, line 13: This type must match the type returned by the endpoints. In this case, it should be `delayedJobs` (make sure to update example in OpenAPI docs as well). IMO it would also make sense to rename the classes to `DelayedJob`Data/Attributes

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -26,13 +26,21 @@ export class DelayedJob extends Model<DelayedJob> {
...
+  processedContent: number | null;
 
   @AllowNull
   @Column(STRING)
-  proccess_message: string | null
+  progressMessage: string | null
+
+  @AllowNull
+  @Column(BIGINT)
+  createdBy: number | null;
```
roguenet, line 41: This should also get a foreign key association, even if we don't necessarily need it right now:
```
  @ForeignKey(() => User)
```

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -26,13 +26,21 @@ export class DelayedJob extends Model<DelayedJob> {
...
-  processed_content: number | null;
+  processedContent: number | null;
 
   @AllowNull
   @Column(STRING)
-  proccess_message: string | null
+  progressMessage: string | null
+
+  @AllowNull
+  @Column(BIGINT)
```
roguenet, line 40: This needs to match what laravel uses, which is `BIGINT.UNSIGNED`

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -1,38 +1,139 @@
...
+      }
+
+      return job;
+    });
+
+    const updatedJobs = await Promise.all(updatePromises);
+
+    return {
+      data: updatedJobs,
+    };
```
roguenet, line 136: Please always use the JsonApiBuilder pattern for responses. You can also either remove the return type from the method, or if you want to keep it, change it to `Promise<JsonApiDocument>`

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -1,38 +1,139 @@
...
   }
-}
+
+  @Patch('bulk-clear')
+  @ApiOperation({
+    operationId: 'bulkClearJobs',
+    summary: 'Bulk update jobs to modify isAcknowledged for specified job IDs',
+    description: `Accepts a JSON:API-compliant payload to bulk update jobs, allowing each job's isAcknowledged attribute to be set to true or false.`,
+  })
+  @ApiBody({
```
roguenet, line 74: Please use the `@JsonApiResponse` decorator for these methods.

roguenet: I see that you added @JsonApiResponse. It should be replacing this one however.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -97,19 +97,23 @@ export class DelayedJobsController {
...
-
+  
     if (!jobUpdates || jobUpdates.length === 0) {
       throw new BadRequestException('No jobs provided in the payload.');
     }
-
+  
+    if (!authenticatedUserId) {
+      throw new UnauthorizedException('Authentication failed.');
+    }
```
roguenet, line 116: This is done automatically by middleware by default. FYI for the future, you have to use @NoBearerAuth to disable it.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -97,19 +97,23 @@ export class DelayedJobsController {
...
+    @Body() bulkClearJobsDto: DelayedJobBulkUpdateBodyDto,
     @Request() { authenticatedUserId }
-  ): Promise<{ data: JobData[] }> {
+  ): Promise<JsonApiDocument> {
     const jobUpdates = bulkClearJobsDto.data;
-
+  
     if (!jobUpdates || jobUpdates.length === 0) {
       throw new BadRequestException('No jobs provided in the payload.');
     }
```
roguenet, line 112: This should be handled by validation decorators in the request DTO. See the bulk update polygon indicator endpoint for examples. In general, we have tools in place to limit the amount of manual checking needed within controller methods. Ideally, by the time the request makes it to the controller, the inputs have been more or less validated and don't need further checking.

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: Added

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: changed

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: Changed to correct pattern

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: Now I think is consistent and returning the delayedJobDto /

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: removed

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: removed too

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -1,38 +1,116 @@
...
+      if (updatedCount === 0) {
+        throw new NotFoundException(`Job with UUID ${job.uuid} could not be updated.`);
+      }
+    
+      const updatedJob = await DelayedJob.findOne({
+        where: { uuid: job.uuid },
+      });
+
+      return updatedJob;
+    });
```
roguenet, line 103: I see how you got here, but it's a little weird to update and then fetch. Let's change that around, reducing the number of queries in the process, and preventing a partial update if some of the jobs aren't found:
```
const jobs = await DelayedJob.findAll({ 
  where: {
    uuid: { [Op.in]: jobUpdates.map(({ uuid }) => uuid) },
    createdBy: authenticatedUserId,
    status: { [Op.ne]: 'pending' }
  }
});
if (jobs.length !== jobUpdates.length) {
  throw new NotFoundException('Some jobs in the request could not be updated');
}

const updatePromises = jobUpdates.map(async ({ uuid, attributes }) => {
  const job = jobs.find((job) => job.uuid === uuid);
  job.isAcknowledged = attributes.isAcknowledged;
  await job.save();
  return job;
});
```

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -1,38 +1,116 @@
...
+    const updatedJobs = await Promise.all(updatePromises);
+    
+    
+    const jsonApiBuilder = buildJsonApi();
+    updatedJobs.forEach((job) => {
+      jsonApiBuilder.addData(job.uuid, new DelayedJobDto(job));
+    });
+    
+    return jsonApiBuilder.serialize();
+
```
roguenet, line 114: :art: It's a minor thing, but these extra blank lines are a little annoying. If you set up your editor to use the prettier config on save, it'll take care of stuff like that for you.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -1,38 +1,116 @@
...
-
     return buildJsonApi()
       .addData(pathUUID, new DelayedJobDto(job))
       .document.serialize();
   }
-}
+
+  @Patch('bulk-clear')
+  @ApiOperation({
+    operationId: 'bulkClearJobs',
```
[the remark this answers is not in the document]
egrojMonroy: renamed

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: great, that prettier will be helpfull. Changed

roguenet (approved):

PR #31, opened 2024-12-17
[TM-1531] Add site name in entity and return entityname in delayed job

egrojMonroy:


libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -1,7 +1,24 @@
...
+  PrimaryKey,
+  Table
+} from "sequelize-typescript";
 import { BIGINT, BOOLEAN, INTEGER, JSON, STRING, UUID } from "sequelize";
 import { User } from "./user.entity";
+import { Site } from "./site.entity";
+import { Project } from "./project.entity";
+import { Logger } from "@nestjs/common";
 
+const ENTITY_TYPE_MAPPING: Record<string, any> = {
```
roguenet, line 18: We try very hard to avoid `any` in this codebase. I know this kind of map is hard to create with correct types... fortunately, there's a better solution outlined below.

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -44,5 +61,29 @@ export class DelayedJob extends Model<DelayedJob> {
...
+   */
+  async getRelatedEntity() {
+    const Model = ENTITY_TYPE_MAPPING[this.entityType || ""];
+    if (Model && this.entityId) {
+      const entity = await Model.findByPk(this.entityId);
+      return entity?.name;
+    }
+
+    return null;
+  }
```
roguenet, line 88: This related entity system should be using scoped `@BelongsTo` associations (see tree-species.entity.ts for an example). With that in place, it should be possible to update the query in the controller to fetch the associated record at the same time as the jobs by including the association. Let me know if you have trouble wiring that up; I've spent a fair mount of time messing around with those systems at this point.

apps/job-service/src/jobs/dto/delayed-job.dto.ts
```diff
@@ -55,4 +55,17 @@ export class DelayedJobDto extends JsonApiAttributes<DelayedJobDto> {
...
+    nullable: true
+  })
+  name: string | null;
+
+  @ApiProperty({
+    description: "The name of the related entity (e.g., site, project).",
+    nullable: true,
+    required: false
+  })
+  entityName?: string | null;
```
roguenet, line 70: I think the entity type should be included here too, even if it's not needed for the current feature.

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: Removed the use of entity type mapping

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: changed to belong to

apps/job-service/src/jobs/dto/delayed-job.dto.ts
egrojMonroy: added

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -44,5 +62,56 @@ export class DelayedJob extends Model<DelayedJob> {
...
+    constraints: false,
+    scope: {
+      entityType: "App\\Models\\V2\\Sites\\Site"
+    },
+    as: "entitySite"
+  })
+  entitySite?: Site;
+
+  declare getEntityProject: () => Promise<Project | null>;
+  declare getEntitySite: () => Promise<Site | null>;
```
roguenet, line 100: I don't think these are super useful. With this TS extension, `this.$get('entityProject')` comes out typed correctly.

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -44,5 +62,56 @@ export class DelayedJob extends Model<DelayedJob> {
...
+
+  async getRelatedEntity(): Promise<string | null> {
+    if (!this.entityId) return null;
+
+    if (this.entityType === "App\\Models\\V2\\Projects\\Project") {
+      const project = this.entityProject ?? (await this.getEntityProject());
+      return project?.name ?? null;
```
roguenet, line 107: If you're needing to load the relation, it's best to cache it on this instance so that a further process in this thread doesn't have to do it again. I would set these lines as 
```
this.entityProject ??= await this.$get('entityProject');
return this.entityProject?.name ?? null:
```

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -39,8 +39,15 @@ export class DelayedJobsController {
       order: [["createdAt", "DESC"]]
     });
```
roguenet, line 40: It would be best to use includes here to avoid extra queries. In the `findAll`:
```
include: [
  { association: "entityProject" },
  { association: "entitySite" }
]
```

For bonus points you can limit the columns that are fetched by adding `attributes: ["name"]` to the include.

In your local env, please make sure it's using a `LEFT OUTER JOIN` for those includes. I _think_ it switches to an `INNER JOIN` when you add a `where` to the include (based on something I was seeing on my most recent v3 branch), so you should be good to go, but if it's using an inner join, you can add `required: false` to the include to get it to switch to a left outer join.

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -1,9 +1,27 @@
...
+} from "sequelize-typescript";
 import { BIGINT, BOOLEAN, INTEGER, JSON, STRING, UUID } from "sequelize";
 import { User } from "./user.entity";
+import { Site } from "./site.entity";
+import { Project } from "./project.entity";
+
+interface EntityAssociations {
+  getEntityProject(): Promise<Project | null>;
+  getEntitySite(): Promise<Site | null>;
+}
```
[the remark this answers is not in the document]
egrojMonroy: removed

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -44,5 +62,56 @@ export class DelayedJob extends Model<DelayedJob> {
...
+  @Column
+  entityId: number;
+
+  @BelongsTo(() => Project, {
+    foreignKey: "entityId",
+    constraints: false,
+    scope: {
+      entityType: "App\\Models\\V2\\Projects\\Project"
+    },
+    as: "entityProject"
```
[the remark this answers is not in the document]
egrojMonroy: also removed. 
Removed the entityType too, because it was generating conflict with:
include: [
  { association: "entityProject" },
  { association: "entitySite" }
]

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: removed

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: changed, nonetheless couldnt use you code exactly because there was an error that i couldnt fix it.
Even though is good.

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: added this includes

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -97,7 +97,20 @@ export class DelayedJobsController {
...
+          attributes: ["name"],
+          required: false
+        },
+        {
+          association: "entitySite",
+          attributes: ["name"],
+          required: false
+        }
+      ],
+      logging: console.log
```
roguenet, line 113: This should not be necessary. At least in both my local and AWS, query logging is already functional.

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -78,37 +73,26 @@ export class DelayedJob extends Model<DelayedJob> implements EntityAssociations
...
   entityProject?: Project;
 
   @BelongsTo(() => Site, {
     foreignKey: "entityId",
-    constraints: false,
-    scope: {
-      entityType: "App\\Models\\V2\\Sites\\Site"
-    },
-    as: "entitySite"
+    constraints: false
```
roguenet, line 82: Why was the `entityType` scope removed? That, I believe, is required for this to function correctly.

libs/database/src/lib/entities/delayed-job.entity.ts
roguenet: Hmm, maybe the `??=` operator isn't set up in this codebase? If that's not working, then this code should still do it the more verbose way:
```
if (this.entityProject == null) {
  this.entityProject = await this.$get("entityProject");
}
return this.entityProject?.name;
```

libs/database/src/lib/entities/delayed-job.entity.ts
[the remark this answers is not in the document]
roguenet: Ah, I missed this comment when I did my most recent code review pass. This is definitely concerning. Can you share with me the details of the conflict and/or the DB query that's being issued? I'm doing something very similar in the most recent iteration of the entity service. It's a little difficult to follow, but [this statement](https://github.com/wri/terramatch-microservices/blob/release/utltimate-ulmus/apps/entity-service/src/trees/tree.service.ts#L73), in the case of a `Site`, is effectively creating two included associations using the same pattern I was outlining here (both joins against a single table, but with different scopes).

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -90,25 +96,34 @@ export class DelayedJobsController {
         uuid: { [Op.in]: jobUpdates.map(({ uuid }) => uuid) },
         createdBy: authenticatedUserId,
         status: { [Op.ne]: "pending" }
-      }
+      },
+      logging: console.log,
```
roguenet, line 100: This shouldn't be needed on these queries, and should not go to production.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -90,25 +96,34 @@ export class DelayedJobsController {
...
+      order: [["createdAt", "DESC"]]
     });
-    if (jobs.length !== jobUpdates.length) {
+
+    if (!jobs.length) {
```
roguenet, line 104: I'm curious about this change; shouldn't this still compare against the length of jobs in the request body?

libs/database/src/lib/entities/delayed-job.entity.ts
```diff
@@ -1,7 +1,22 @@
...
+  Table
+} from "sequelize-typescript";
 import { BIGINT, BOOLEAN, INTEGER, JSON, STRING, UUID } from "sequelize";
 import { User } from "./user.entity";
 
+interface JobMetadata {
+  entity_id: string; // or number, depending on your model's entity ID type
+  entity_type: string;
+  entity_name: string; // Name of the related entity
+}
```
roguenet, line 19: I don't think we should try to strongly type this. Future jobs could put anything they want in `metadata`, and we wouldn't want to have to update the schema here each time that happens. The `payload` below simply uses `object`, and that should work fine for this aw well.

libs/database/src/lib/entities/site-name.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { AutoIncrement, Column, Model, PrimaryKey, Table } from "sequelize-typescript";
+import { BIGINT } from "sequelize";
+
+// A quick stub for the research endpoints
+@Table({ tableName: "v2_sites", underscored: true, paranoid: true })
```
roguenet, line 5: We already have a `site.entity.ts` that reflects this table. I can't think of any reason why we'd want to have two, even if sequelize is able to handle that correctly (I suspect we'd run into problems when v3 eventually becomes the source of truth for the table schemas and is responsible for migrations)

egrojMonroy: sorry, that file shouldnt reach here, it was one of my tests to try an entity with only name for the query. 
Deleted

libs/database/src/lib/entities/delayed-job.entity.ts
egrojMonroy: changed to object

apps/job-service/src/jobs/delayed-jobs.controller.ts
roguenet: Hmm, but this code as written won't detect if the FE tried to update a job that the user doesn't own. It would be better for the FE to only send the IDs that it expects to be able to update (so, filtering out pending ones) so that the BE can raise the issue correctly. I would certainly consider it an error that the BE should complain about if the FE asks it to update job that can't be updated (because it's in pending, or because it's owned by another user doesn't make much difference; it's a breaking in the contract of the endpoint to not update something that was asked to be updated and also not complain about it)

egrojMonroy: this has being changed. Now The FE will only send the jobs that will be updated.

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: the logs is removed

apps/job-service/src/jobs/delayed-jobs.controller.ts
egrojMonroy: reverted the change too

roguenet (approved):

PR #35, opened 2024-12-26
[TM-1531] prevent error on metadata null

egrojMonroy:
Error was triggering on trying to get entity_name from metadata

### DIFF
--- a/apps/job-service/src/jobs/delayed-jobs.controller.ts
+++ b/apps/job-service/src/jobs/delayed-jobs.controller.ts
@@ -41,7 +41,7 @@ export class DelayedJobsController {
 
     const jobsWithEntityNames = await Promise.all(
       runningJobs.map(async job => {
-        const entityName = job.metadata ? (job.metadata as { entity_name?: string }).entity_name : null;
+        const entityName = job.metadata?.entity_name;
         return { ...job.toJSON(), entityName };
       })
     );
@@ -111,7 +111,7 @@ export class DelayedJobsController {
         job.isAcknowledged = attributes.isAcknowledged;
         await job.save();
 
-        const entityName = job.metadata ? (job.metadata as { entity_name?: string }).entity_name : null;
+        const entityName = job.metadata?.entity_name;
 
         return { ...job.toJSON(), entityName };
       });
--- a/apps/job-service/src/jobs/delayed-jobs.controller.ts
+++ b/apps/job-service/src/jobs/delayed-jobs.controller.ts
@@ -41,7 +41,7 @@ export class DelayedJobsController {
 
     const jobsWithEntityNames = await Promise.all(
       runningJobs.map(async job => {
-        const entityName = (job.metadata as { entity_name?: string }).entity_name;
+        const entityName = job.metadata ? (job.metadata as { entity_name?: string }).entity_name : null;
         return { ...job.toJSON(), entityName };
       })
     );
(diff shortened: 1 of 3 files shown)


apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -111,7 +111,7 @@ export class DelayedJobsController {
         job.isAcknowledged = attributes.isAcknowledged;
         await job.save();
 
-        const entityName = (job.metadata as { entity_name?: string }).entity_name;
+        const entityName = job.metadata ? (job.metadata as { entity_name?: string }).entity_name : null;
```
roguenet, line 114: These should not be relying on any particular structure to the metadata. I think this would be cleaner:
```
const entityName = job.metadata?.entity_name ?? "";
```

egrojMonroy: I couldn't use it that way because I had this error: 
Property 'entity_name' does not exist on type 'object'.ts(2339)

Because the definition is like this:
```
 @AllowNull
  @Column(JSON)
  metadata: object | null;
```
Some options to fix it:
  
`  const entityName = (job.metadata as any)?.entity_name ?? "";`

or:
```
@AllowNull
@Column(JSON)
metadata: any | null;
```

roguenet: Yeah, it's pretty awkward dealing with `object` when we want to avoid `any`. This works:
```
        const entityName = ((job.metadata ?? {}) as { entity_name?: string }).entity_name;
```

If we end up needing to do a lot of juggling with metadata members, it _might_ be worth having some typed shapes of expected metadata and use some type narrowing predicates to make it simpler, but I think the above is fine for now.

roguenet: This is also fine (please do include the comment if you go this route; I want it to be clear that the interface isn't meant to be a complete declaration of all metadata members on every job
```
diff --git a/apps/job-service/src/jobs/delayed-jobs.controller.ts b/apps/job-service/src/jobs/delayed-jobs.controller.ts
index d137848..f9ee12a 100644
--- a/apps/job-service/src/jobs/delayed-jobs.controller.ts
+++ b/apps/job-service/src/jobs/delayed-jobs.controller.ts
@@ -111,7 +111,7 @@ export class DelayedJobsController {
         job.isAcknowledged = attributes.isAcknowledged;
         await job.save();
 
-        const entityName = (job.metadata as { entity_name?: string }).entity_name;
+        const entityName = job.metadata?.entity_name;
 
         return { ...job.toJSON(), entityName };
       });
diff --git a/libs/database/src/lib/entities/delayed-job.entity.ts b/libs/database/src/lib/entities/delayed-job.entity.ts
index 772288e..e76a0e2 100644
--- a/libs/database/src/lib/entities/delayed-job.entity.ts
+++ b/libs/database/src/lib/entities/delayed-job.entity.ts
@@ -12,6 +12,12 @@ import {
 import { BIGINT, BOOLEAN, INTEGER, JSON, STRING, UUID } from "sequelize";
 import { User } from "./user.entity";
 
+// holds the definition for members that may exist in a job metadata that this codebase explicitly
+// references.
+interface Metadata {
+  entity_name?: string;
+}
+
 @Table({ tableName: "delayed_jobs", underscored: true })
 export class DelayedJob extends Model<DelayedJob> {
   @PrimaryKey
@@ -61,5 +67,5 @@ export class DelayedJob extends Model<DelayedJob> {
 
   @AllowNull
   @Column(JSON)
-  metadata: object | null;
+  metadata: Metadata | null;
 }
```

egrojMonroy: The second option is ready in the code.

roguenet (approved):

PR #42, opened 2025-01-16
[TM-1522] Daily syncs of the DB to Airtable.

roguenet:
https://gfw.atlassian.net/browse/TM-1522

Also included: 
* Slack update messages as processing takes place or errors out
* Health check on the redis connection

### DIFF
--- a/apps/unified-database-service/src/airtable/airtable.processor.spec.ts
+++ b/apps/unified-database-service/src/airtable/airtable.processor.spec.ts
@@ -4,6 +4,7 @@ import { ConfigService } from "@nestjs/config";
 import { createMock } from "@golevelup/ts-jest";
 import { InternalServerErrorException, NotImplementedException } from "@nestjs/common";
 import { Job } from "bullmq";
+import { SlackService } from "nestjs-slack";
 
 jest.mock("airtable", () =>
   jest.fn(() => ({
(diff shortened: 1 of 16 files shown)


apps/unified-database-service/src/airtable/airtable.processor.spec.ts
```diff
@@ -66,4 +71,27 @@ describe("AirtableProcessor", () => {
...
+        jest.spyOn(processor as never, "updateEntities") as jest.SpyInstance<Promise<void>>
+      ).mockImplementation(() => Promise.resolve());
+      const deleteSpy = (
+        jest.spyOn(processor as never, "deleteEntities") as jest.SpyInstance<Promise<void>>
+      ).mockImplementation(() => Promise.resolve());
+
+      const updatedSince = new Date();
+      await processor.process({ name: "updateAll", data: { updatedSince } } as Job);
+
+      for (let ii = 0; ii < ENTITY_TYPES.length; ii++) {
```
pachonjcl, line 87: seeing the `ii` instead of `i` or `j` blow my mind.

pachonjcl (approved):

apps/unified-database-service/src/airtable/airtable.processor.spec.ts
roguenet: It's a habit I got in to at the beginning of my career that doesn't super make sense anymore, but I still do it. Back then, tools for finding uses of a variable weren't nearly as good, so a senior engineer at the time convinced me to use `ii` instead of `i` because when you use a tool like grep or regex in vim to search through code, `i` is all over the place, but `ii` is distinct and much easier to find.

PR #46, opened 2025-01-22
[TM-1311] Implement password reset 

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1311


apps/user-service/src/auth/reset-password.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+  })
+  @JsonApiResponse({ status: HttpStatus.OK, data: { type: ResetPasswordResponseOperationDto } })
+  @ApiException(() => BadRequestException, { description: 'Invalid or expired token.' })
+  async resetPassword(
+    @Param('token') token: string,
+    @Body() { newPassword }: ResetPasswordDto
+  ): Promise<JsonApiDocument> {
+    const response = await this.resetPasswordService.resetPassword(token, newPassword);
+    return buildJsonApi()
+      .addData('sads',response)
```
roguenet, line 52: `'sads'` - is this a testing / debug value? In keeping with the comment above, the ID used for each of these endpoints in its response should also match. That's important for the FE caching layer. In this case I would use the user id for all of them to match the login controller.

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -0,0 +1,72 @@
...
+import {ConfigService} from "@nestjs/config";
+import { ResetPasswordResponseOperationDto } from "./dto/reset-password-response-operation.dto";
+
+@Injectable()
+export class ResetPasswordService {
+  constructor(
+    private readonly jwtService: JwtService,
+    private readonly emailService: EmailService,
+    private readonly configService: ConfigService,
+    //private readonly userService: UserService  // Assuming you have a User service to interact with the database
```
roguenet, line 18: Please don't leave in commented out code.

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async sendResetPasswordEmail(emailAddress: string) {
+    const user = await User.findOne({ where: { emailAddress } });
+    if (!user) {
+      throw new NotFoundException('User not found');
+    }
+
+    const resetToken = await this.jwtService.signAsync(
+      { sub: user.uuid }, // user id as the subject
+      { expiresIn: '1h', secret: 'reset_password_secret' } // token expires in 1 hour
```
roguenet, line 29: What is this secret accomplishing? Since our JWTs are signed with a secret key, I don't think there's any real value here. If there is some value to including it, it should be from the environment (and be a random string), not included in our publicly accessible codebase.

libs/database/src/lib/entities/localization-keys.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  @AutoIncrement
+  @Column(BIGINT.UNSIGNED)
+  override id: number;
+
+  @AllowNull
+  @Column(STRING)
+  key: string | null;
+
+  @AllowNull
+  @Column(STRING)
```
roguenet, line 27: This column appears to be a TEXT type, not STRING in the DB (`text` vs `varchar(255)`)

libs/database/src/lib/entities/localization-keys.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  value: string | null;
+
+  @AllowNull
+  @Unique
```
roguenet, line 31: I'm not sure if these value ids are required to be unique or not, but I do see that the current DB schema doesn't set a uniqueness constraint on this column, so we shouldn't here either.

roguenet (commented):
I don't see any specs in this PR. Please ensure you have specs for all new controllers and services that pass our global coverage configuration.

apps/user-service/src/auth/dto/reset-password-response.dto.ts
```diff
@@ -0,0 +1,19 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { JsonApiAttributes } from "@terramatch-microservices/common/dto/json-api-attributes";
+
+@JsonApiDto({ type: 'logins', id: 'number' })
```
roguenet, line 5: This is not a login. Please choose a different resource type. It should likely match the endpoints. In this case, `passwordResets` would make sense to me.

apps/user-service/src/auth/dto/reset-password-response.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+export class ResetPasswordResponseDto extends JsonApiAttributes<ResetPasswordResponseDto> {
+  @ApiProperty({
+    description: 'User id',
```
roguenet, line 8: Let's be careful with language around "id"s. This is a UUID, not an ID (IDs are numbers)

apps/user-service/src/auth/reset-password.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+  @ApiOperation({
+    operationId: 'requestPasswordReset',
+    description: 'Send password reset email with a token',
+  })
+  @JsonApiResponse({ status: HttpStatus.CREATED, data: { type: ResetPasswordResponseDto } })
+  @ApiException(() => BadRequestException, { description: 'Invalid request or email.' })
+  async requestReset(@Body() { emailAddress, callbackUrl }: ResetPasswordRequest): Promise<JsonApiDocument> {
+    const { email, userId, uuid } = await this.resetPasswordService.sendResetPasswordEmail(emailAddress, callbackUrl);
+    return buildJsonApi()
+      .addData(`${userId}`, new ResetPasswordResponseDto({ emailAddress: email, uuid }))
```
roguenet, line 34: Why is both the user id and the user UUID needed? Maybe I'll see it when I look at the FE integration, but I suspect this would do better putting the UUID in as the resource ID and leaving the user ID out of it.

libs/common/src/lib/email/email.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Injectable()
+export class EmailService {
+  private transporter: nodemailer.Transporter;
+
+  constructor(private readonly configService: ConfigService) {
+    this.transporter = nodemailer.createTransport({
+      host: this.configService.get<string>('MAIL_HOST'),
+      port: this.configService.get<number>('MAIL_PORT'),
+      secure: false, // true for 465, false for other ports
```
roguenet, line 13: This comment is a little odd. What if `MAIL_PORT` where configured to be 465? What will it be in AWS?

libs/common/src/lib/localization/localization.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        [Op.or]: [
+          { shortValue: value },
+          { longValue: value }
+        ]
+      }
+    });
+  }
+
+  private async getTranslateItem(itemId: number, locale: string): Promise<i18nTranslation | null> {
+    return await i18nTranslation.findOne({where: { i18nItemId: itemId, language: locale}});
```
roguenet, line 26: There are several cases of our `prettier` styles not being applied in this PR. Please set that up locally and apply it. Your editor should be able to automatically have prettier run on file save. Some time soon I'll get our linter set up to fail on things like this (the spacing around your curly braces is missing). I don't really understand why the out of the box NX / NestJS setup doesn't enforce that already.

libs/common/src/lib/localization/translation.service.ts
```diff
@@ -0,0 +1,29 @@
+import { Injectable } from '@nestjs/common';
+import { tx, t, normalizeLocale } from "@transifex/native";
+import { ConfigService } from "@nestjs/config";
+
+@Injectable()
+export class TranslationService {
+  constructor(private readonly configService: ConfigService,) {
```
roguenet, line 7: If the service isn't going to be used after the class is constructed, there's no reason to make it a property of the class, so this can be simpler:
```
constructor(configService: ConfigService) {
```

libs/common/src/lib/localization/translation.service.ts
```diff
@@ -0,0 +1,29 @@
+import { Injectable } from '@nestjs/common';
+import { tx, t, normalizeLocale } from "@transifex/native";
+import { ConfigService } from "@nestjs/config";
+
+@Injectable()
+export class TranslationService {
```
roguenet, line 6: I'm not sure I see value in keeping the translation service and localization service separate. They're both tiny, and they're used for similar things. I think future developers wouldn't necessarily know where new code is supposed to go between these two, as the line is often blurry.

libs/database/src/lib/entities/i18n-item.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  @AutoIncrement
+  @Column(BIGINT.UNSIGNED)
+  override id: number;
+
+  @AllowNull
+  @Column(STRING)
+  status: string | null;
+
+  @AllowNull
+  @Column(STRING)
```
roguenet, line 24: This column has a default value of `short` at the DB level. Please reflect that in the column definition.

libs/database/src/lib/entities/i18n-item.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  type: string | null;
+
+  @AllowNull
+  @Column(STRING)
+  shortValue: string | null;
+
+  @AllowNull
+  @Column(STRING)
```
roguenet, line 32: Looking at the current schema, this should be a `TEXT`, not a `STRING`. We need to keep these types as close to what's in the current schema as possible because this codebase will eventually become the source of truth for what the schema in the prod DB will be.

libs/database/src/lib/entities/i18n-item.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  AllowNull,
+  AutoIncrement,
+  Column,
+  Model,
+  PrimaryKey,
+  Table,
+} from "sequelize-typescript";
+import { BIGINT, STRING } from "sequelize";
+
+@Table({ tableName: "i18n_items", underscored: true, paranoid: false })
```
roguenet, line 11: `paranoid` tells sequelize to add/manage a `deleted_at` field, which this table doesn't have. Please remove `paranoid` from the table def here.

libs/database/src/lib/entities/i18n-translation.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+  AllowNull,
+  AutoIncrement,
+  Column,
+  Model,
+  PrimaryKey,
+  Table, Unique
+} from "sequelize-typescript";
+import { BIGINT, INTEGER, NUMBER, STRING } from "sequelize";
+
+@Table({ tableName: "i18n_translations", underscored: true, paranoid: false })
```
roguenet, line 11: This table is also not paranoid

libs/database/src/lib/entities/i18n-translation.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+  @AllowNull
+  @Column(STRING)
+  language: string | null;
+
+  @AllowNull
+  @Column(STRING)
+  shortValue: string | null;
+
+  @AllowNull
+  @Column(STRING)
```
roguenet, line 33: This one should also be `TEXT`

libs/database/src/lib/entities/index.ts
```diff
@@ -33,3 +33,6 @@ export * from "./tree-species.entity";
 export * from "./tree-species-research.entity";
 export * from "./user.entity";
 export * from "./workday.entity";
+export * from "./localization-keys.entity";
+export * from "./i18n-item.entity";
+export * from "./i18n-translation.entity";
```
roguenet, line 38: Please keep this file organized alphabetically.

libs/database/src/lib/entities/localization-keys.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+export class LocalizationKey extends Model<LocalizationKey> {
+
+  @PrimaryKey
+  @AutoIncrement
+  @Column(BIGINT.UNSIGNED)
+  override id: number;
+
+  @NotNull
+  @AllowNull(false)
+  @Column(TEXT)
```
roguenet, line 22: This should be `STRING` (varchar(255) in the DB)

libs/database/src/lib/entities/localization-keys.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+@Table({ tableName: "localization_keys", underscored: true })
+export class LocalizationKey extends Model<LocalizationKey> {
+
+  @PrimaryKey
+  @AutoIncrement
+  @Column(BIGINT.UNSIGNED)
+  override id: number;
+
+  @NotNull
+  @AllowNull(false)
```
roguenet, line 21: Why are both of these on here? I'm pretty sure not null is the default, so these may both be removed. Also applies to `value` below. Finally, if a column is not nullable, the TS type of the field should not include `null`. So this one should be:
```
@Column(STRING)
key: string;
```

libs/database/src/lib/entities/localization-keys.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+  @Column(TEXT)
+  key: string | null;
+
+  @NotNull
+  @AllowNull(false)
+  @Column(TEXT)
+  value: string | null;
```
roguenet, line 28: As mentioned above, this should simply be
```
@Column(TEXT)
value: string;
```

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -0,0 +1,81 @@
...
+@Injectable()
+export class ResetPasswordService {
+  private readonly logger: Logger;
+
+  constructor(
+    private readonly jwtService: JwtService,
+    private readonly emailService: EmailService,
+    private readonly localizationService: LocalizationService
+  ) {
+    this.logger = new Logger(ResetPasswordService.name);
```
roguenet, line 17: Please use `TMLogService`

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  async sendResetPasswordEmail(emailAddress: string, callbackUrl: string) {
+    const user = await User.findOne({ where: { emailAddress } });
```
roguenet, line 21: It looks like you're only using id, uuid and locale from the user so I'd limit this query to those attributes:
```
const user = await User.findOne({ where: { emailAddress }, attributes: ["id", "uuid", "locale"]);
```

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -0,0 +1,81 @@
...
+    const resetLink = `${callbackUrl}/${resetToken}`;
+    const bodyEmail = this.formatBody(bodyEmailContent, resetLink);
+    await this.emailService.sendEmail(user.emailAddress, subjectLocalization.value, bodyEmail);
+
+    return { email: user.emailAddress, uuid: user.uuid, userId: user.id };
+  }
+
+  private formatBody(bodyEmailContent: string, resetLink: string) {
+    const anchor = `<a href="${resetLink}" target="_blank">link</a>`;
+    return bodyEmailContent.replace('link', anchor).replace('enlace', anchor).replace('lien', anchor);
```
roguenet, line 58: Oof... not something that we need to fix in this PR, but the replacement text should not be translated. A developer shouldn't need to know what "link" is in all of our translated languages. Plus, this will break when we add another one.  In fact, does this cover Portuguese? I only see EN, ES and FR represented here.

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -0,0 +1,81 @@
...
+    let userGuid;
+    try {
+      const payload = await this.jwtService.verifyAsync(resetToken);
+      userGuid = payload.sub;
+    } catch (error) {
+      this.logger.error(error);
+      throw new BadRequestException('Provided token is invalid or expired');
+    }
+
+    const user = await User.findOne({ where: { uuid: userGuid } });
```
roguenet, line 71: I would use `attributes: ["id", "uuid"]` here since you're not using anything else from the model.

libs/common/src/lib/email/email.service.ts
Scriptmatico: This is just for the actual configuration, we can take the config value from the configservice in case that is needed, but for now I think is ok, or do you want me to take this value from env?

libs/common/src/lib/localization/translation.service.ts
Scriptmatico: Do you want me to isolate translations service and localization? Im not sure if is the best approach, but if you see that is needed, well is ok for me.

libs/common/src/lib/email/email.service.ts
roguenet: I was mainly reacting to the comment noting that for some port values this should be true, but the port value is configured externally. I would do this line as `
```
secure: this.configService.get<number>('MAIN_PORT') === 465,
```

apps/user-service/src/auth/reset-password.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+  @ApiOperation({
+    operationId: 'requestPasswordReset',
+    description: 'Send password reset email with a token',
+  })
+  @JsonApiResponse({ status: HttpStatus.CREATED, data: { type: ResetPasswordResponseDto } })
+  @ApiException(() => BadRequestException, { description: 'Invalid request or email.' })
+  async requestReset(@Body() { emailAddress, callbackUrl }: ResetPasswordRequest): Promise<JsonApiDocument> {
+    const { email, userId } = await this.resetPasswordService.sendResetPasswordEmail(emailAddress, callbackUrl);
+    return buildJsonApi()
+      .addData(`${userId}`, new ResetPasswordResponseDto({ emailAddress: email }))
```
roguenet, line 34: Please use UUID here (and update the type on the DTO decorator). We should be favoring UUIDs for all API surface areas as much as possible. 

I realize the login controller is using numerical ID. It doesn't actually matter in that case, and I should probably change (and the user find endpoint; while only used for `me` right now, it does accept a numerical ID for future admin use that I want to switch to UUID). The recently added user update endpoint does use UUID though, and all other endpoints that exist today and refer to a real resource use UUID.

libs/common/src/lib/localization/translation.service.ts
roguenet: If you feel strongly about keeping them separate, you can. If it were me, I would just fold the stuff from Localization service into this service and just have the one service.

libs/database/src/lib/entities/i18n-item.entity.ts
```diff
@@ -21,19 +13,18 @@ export class i18nItem extends Model<i18nItem> {
   status: string | null;
 
   @AllowNull
-  @Column(STRING)
+  @Column(DataTypes.STRING(255))
```
roguenet, line 16: varchar(255) is the default for `STRING`, so adding the 255 here is not needed.

libs/database/src/lib/entities/index.ts
```diff
@@ -1,38 +1,37 @@
 export * from "./application.entity";
 export * from "./delayed-job.entity";
-export * from "./demographic.entity";
 export * from "./demographic-entry.entity";
+export * from "./demographic.entity";
 export * from "./form-submission.entity";
-export * from "./framework.entity";
 export * from "./framework-user.entity";
+export * from "./framework.entity";
```
roguenet, line 7: Not that it's a big deal, but why these changes and the the others here around files that aren't part of this PR? `framework` should come before `framework-user` in an alphabetical sort

libs/database/src/lib/entities/i18n-item.entity.ts
roguenet: I haven't been as strict about getting things perfect as I was originally thinking I would. If it's useful, one tool I've used to check is one you've run some unit tests with the updated schema in place, you can use `describe table` on both the PHP database, and the testing database used for this service. That'll show you easily what differences there are. 

However, I think that when we are ready to make this repo be the owner of the schema (maybe by summer? not sure), we'll probably have to do a big pass over all the tables, use some tooling to find any differences in schema and address each one, so right now it's just about limiting the number of column defs we're going to have to tweak at that time, so I just generally try to get them close and worry about it being perfect later. One thing in particular that this codebase isn't great at right now is keeping up with index definitions. I have a few of them specified, but we'll both need to make sure this codebase is aware of all the ones created by PHP when we switch and also do an audit for places where an index would speed things up and doesn't yet exist.

apps/user-service/src/auth/reset-password.service.ts
```diff
@@ -47,9 +47,10 @@ export class ResetPasswordService {
     const bodyEmailContent = await this.localizationService.translate(bodyLocalization.value, user.locale);
     const resetLink = `${callbackUrl}/${resetToken}`;
     const bodyEmail = this.formatBody(bodyEmailContent, resetLink);
+    console.log(user.emailAddress);
```
roguenet, line 50: Please don't leave debug logging in.

roguenet (approved):

PR #58, opened 2025-02-18
[TM-1676] Implement user creation BE 

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1676


.env.local.sample
```diff
@@ -35,3 +35,5 @@ MAIL_PASSWORD=
 MAIL_FROM_ADDRESS=staging@terramatch.org
 
 TRANSIFEX_TOKEN=
+
+APP_BACKEND_URL=https://new-wri-prod.wri-restoration-marketplace-api.com
```
roguenet, line 39: This file should have a value that makes sense for local development.

apps/user-service/src/users/user-creation.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    const role = request.role;
+    const roleEntity = await Role.findOne({ where: { name: role } });
+
+    if (roleEntity == null) {
+      throw new NotFoundException("Role not found");
+    }
+
+    const userExist = await User.findOne({ where: { emailAddress: request.emailAddress } });
```
roguenet, line 69: This would be more efficient with `count()`: 
```
const userExists = (await User.count({ where: { emailAddress: request.emailAddress } })) !== 0;
```

apps/user-service/src/users/user-creation.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (ctaLocalization == null) {
+      throw new NotFoundException("Localization CTA not found");
+    }
+
+    const role = request.role;
+    const roleEntity = await Role.findOne({ where: { name: role } });
+
+    if (roleEntity == null) {
+      throw new NotFoundException("Role not found");
+    }
```
roguenet, line 67: I see that the DTO only allows a limited set of roles, but considering the security implications if that doesn't work correctly, I'd go ahead and throw an error here too if the role passed to the service is different from one of those three.

apps/user-service/src/users/users.controller.ts
```diff
@@ -14,10 +15,13 @@ import { PolicyService } from "@terramatch-microservices/common";
...
 import { ExceptionResponse, JsonApiResponse } from "@terramatch-microservices/common/decorators";
-import { buildJsonApi, DocumentBuilder } from "@terramatch-microservices/common/util";
+import { buildJsonApi, DocumentBuilder, JsonApiDocument } from "@terramatch-microservices/common/util";
 import { UserUpdateBodyDto } from "./dto/user-update.dto";
+import { NoBearerAuth } from "@terramatch-microservices/common/guards";
+import { UserNewRequest } from "./dto/user-new-request.dto";
+import { UserCreationService } from "./user-creation.service";
 
-const USER_RESPONSE_SHAPE = {
+export const USER_RESPONSE_SHAPE = {
```
roguenet, line 24: Does this need to be exported? I don't think it should be used outside of this file and I don't see any users external to this file in this PR.

.env.local.sample
roguenet: Also, I just found where this is being used. We should go ahead and make this use the API gateway. A change was made to the CDK stack for the gateway recently to send requests for `images/v2` to the PHP BE (https://github.com/wri/terramatch-microservices/blob/staging/cdk/api-gateway/lib/api-gateway-stack.ts#L95C1-L95C87). I think if that is simply changed to `images` (dropping the V2), then this can be `api.terramatch.org` for prod, `api-staging.terramatch.org` for staging, etc. That will probably help keep our emails from getting flagged as spam too, since then the image domain will match the from domain on the email.

roguenet (commented):
Looking pretty good! Just a few minor comments.

.env.local.sample
roguenet: Please go ahead and make this change in this PR.

roguenet (commented):
I think this is ready to go once the change has been made to the CDK stack for the gateway.

roguenet (approved):

PR #60, opened 2025-02-19
[TM-1711] update bulkUpdateJobs to allow updating a single job that i…

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1711)

### DIFF
--- a/apps/job-service/src/jobs/delayed-jobs.controller.spec.ts
+++ b/apps/job-service/src/jobs/delayed-jobs.controller.spec.ts
@@ -247,45 +247,6 @@ describe("DelayedJobsController", () => {
 
     //   await expect(controller.bulkUpdateJobs(payload, request)).rejects.toThrow(NotFoundException);
     // });
-
-    it('should not update jobs with status "pending" if multiple exist', async () => {
-      const authenticatedUserId = 130999;
-
-      const pendingJob1 = await DelayedJob.create({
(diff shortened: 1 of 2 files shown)


roguenet (commented):
This looks OK, but there appears to be some spec updates needed. It's also worth noting that while this may make it look like the job cancelled in the FE, the worker process in the PHP BE will still run to completion, so whatever action the user was trying to take in the FE, it will eventually still try to finish in the BE, so this isn't a real "cancel". I don't know that we have a good way to effectively cancel the BE job though, so I suppose this is the best we can do... I'm curious about the need for this though. Is product aware that this won't cancel the actual work?

apps/job-service/src/jobs/delayed-jobs.controller.spec.ts
```diff
@@ -224,29 +224,29 @@ describe("DelayedJobsController", () => {
...
-            uuid: pendingJob.uuid,
-            attributes: { isAcknowledged: true }
-          }
-        ]
-      };
-      const request = { authenticatedUserId };
-
-      await expect(controller.bulkUpdateJobs(payload, request)).rejects.toThrow(NotFoundException);
-    });
+    // it('should not update jobs with status "pending"', async () => {
```
roguenet, line 227: Please do not merge commented out code.

roguenet (dismissed):

PR #63, opened 2025-02-21
[TM-1665] chage collection tree-planted -> nursery-seedling to projec…

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1665)

### DIFF
--- a/libs/database/src/lib/entities/project-report.entity.ts
+++ b/libs/database/src/lib/entities/project-report.entity.ts
@@ -32,7 +32,7 @@ type ApprovedIdsSubqueryOptions = {
 }))
 @Table({ tableName: "v2_project_reports", underscored: true, paranoid: true })
 export class ProjectReport extends Model<ProjectReport> {
-  static readonly TREE_ASSOCIATIONS = ["treesPlanted"];
+  static readonly TREE_ASSOCIATIONS = ["nurserySeedlings"];
   static readonly PARENT_ID = "projectId";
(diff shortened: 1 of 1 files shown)


LimberHope:
@roguenet , it seems like I need to update the test file, am I correct?

roguenet (approved):

roguenet:
Whoops, yeah it appears I was wrong that "trees-planted" on project reports is unused. This is a much more substantial change: The tree service is fetching establishment species on Project when filling in project report. Now that that has moved to `nursery-seedling`, please ask Ben if we should also be changing the species on Project to `nursery-seedling`, because it seems likely that we should, and then this is going to require a bunch of changes to the tree service, and possibly the FE too (although maybe not needed there)

LimberHope:
@roguenet , I checked with Ben, and he’s not sure we should make this change. Since projects are also reporting in site reports using the tree-planted collection type, he asked if we could keep the establishment species on the project as tree-planted, or if all forms need to follow the same collection type. What do you think?

roguenet:
> @roguenet , I checked with Ben, and he’s not sure we should make this change. Since projects are also reporting in site reports using the tree-planted collection type, he asked if we could keep the establishment species on the project as tree-planted, or if all forms need to follow the same collection type. What do you think?

Started a thread in slack to get to the bottom of this.

roguenet:
Moving this to a draft. I'll be taking this ticket over, but am not picking it up quite yet.

PR #64, opened 2025-02-24
[TM-1766] add initial changes for supporting entity get

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1766)

### DIFF
--- a/apps/entity-service/src/entities/processors/site.processor.ts
+++ b/apps/entity-service/src/entities/processors/site.processor.ts
@@ -1,21 +1,31 @@
+import { DocumentBuilder } from "@terramatch-microservices/common/util";
 import { EntityProcessor, PaginatedResult } from "./entity-processor";
-import { Site } from "@terramatch-microservices/database/entities";
-import { SiteFullDto, SiteLightDto } from "../dto/site.dto";
+import { Nursery, ProjectReport, Site } from "@terramatch-microservices/database/entities";
+import { AdditionalSiteFullProps, SiteFullDto, SiteLightDto } from "../dto/site.dto";
 
 export class SiteProcessor extends EntityProcessor<Site, SiteLightDto, SiteFullDto> {
   readonly LIGHT_DTO = SiteLightDto;
   readonly FULL_DTO = SiteFullDto;
 
-  async findOne(): Promise<Site> {
-    throw new Error("Method not implemented.");
+  async findOne(uuid: string): Promise<Site> {
+    return await Site.findOne({
+      where: { uuid },
+      include: [{ association: "framework" }]
+    });
   }
 
   async findMany(): Promise<PaginatedResult<Site>> {
     throw new Error("Method not implemented.");
   }
 
-  async addFullDto(): Promise<void> {
-    throw new Error("Method not implemented.");
+  async addFullDto(document: DocumentBuilder, site: Site): Promise<void> {
+    const siteId = site.id;
+
+    const props: AdditionalSiteFullProps = {
+      totalSiteReports: 0
+    };
+
+    document.addData(site.uuid, new SiteFullDto(site, props));
   }
 
   async addLightDto(): Promise<void> {
--- a/libs/common/src/lib/policies/site.policy.ts
+++ b/libs/common/src/lib/policies/site.policy.ts
@@ -1,0 +1,29 @@
+import { EntityPolicy } from "./entity.policy";
+import { Site, User } from "@terramatch-microservices/database/entities";
+import { FrameworkKey } from "@terramatch-microservices/database/constants/framework";
+
+export class SitePolicy extends EntityPolicy {
+  async addRules() {
+    if (this.permissions.includes("sites-read") || this.permissions.includes("view-dashboard")) {
+      this.builder.can("read", Site);
+    }
+
+    const frameworks = this.permissions
+      .filter(name => name.startsWith("framework-"))
+      .map(name => name.substring("framework-".length) as FrameworkKey);
+    if (frameworks.length > 0) {
+      this.builder.can("read", Site, { frameworkKey: { $in: frameworks } });
+    }
+  }
+
+  protected _user?: User | null;
+  protected async getUser() {
+    if (this._user != null) return this._user;
+
+    return (this._user = await User.findOne({
+      where: { id: this.userId },
+      attributes: ["organisationId"],
+      include: [{ association: "sites", attributes: ["id"] }]
+    }));
+  }
+}
--- a/apps/entity-service/src/entities/dto/site.ts
+++ b/apps/entity-service/src/entities/dto/site.ts
@@ -1,0 +1,21 @@
+import { JsonApiDto } from "@terramatch-microservices/common/decorators/json-api-dto.decorator";
+import { EntityDto } from "./entity.dto";
+import { Site } from "@terramatch-microservices/database/entities";
+import { pickApiProperties } from "@terramatch-microservices/common/dto/json-api-attributes";
+import { ApiProperty } from "@nestjs/swagger";
+
+@JsonApiDto({ type: "sites" })
+export class SiteLightDto extends EntityDto {
+  constructor(site?: Site) {
+    super();
+    if (site != null) {
+      this.populate(SiteLightDto, {
+        ...pickApiProperties(site, SiteLightDto),
+        lightResource: true
+      });
+    }
+  }
+
+  @ApiProperty({ nullable: true, description: "Framework key for this site" })
+  frameworkKey: string | null;
+}
--- a/libs/common/src/lib/policies/policy.service.ts
+++ b/libs/common/src/lib/policies/policy.service.ts
@@ -2,13 +2,14 @@ import { Injectable, LoggerService, Scope, UnauthorizedException } from "@nestjs
 import { RequestContext } from "nestjs-request-context";
 import { UserPolicy } from "./user.policy";
 import { BuilderType, EntityPolicy } from "./entity.policy";
-import { Permission, Project, SitePolygon, User } from "@terramatch-microservices/database/entities";
+import { Permission, Project, Site, SitePolygon, User } from "@terramatch-microservices/database/entities";
 import { AbilityBuilder, createMongoAbility } from "@casl/ability";
 import { Model } from "sequelize-typescript";
 import { TMLogService } from "../util/tm-log.service";
 import { SitePolygonPolicy } from "./site-polygon.policy";
 import { ProjectPolicy } from "./project.policy";
 import { isArray } from "lodash";
+import { SitePolicy } from "./site.policy";
 
 type EntityClass = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -22,6 +23,7 @@ type PolicyClass = {
 
 const POLICIES: [EntityClass, PolicyClass][] = [
   [Project, ProjectPolicy],
+  [Site, SitePolicy],
   [SitePolygon, SitePolygonPolicy],
   [User, UserPolicy]
 ];


pachonjcl:
Will reduce coverage temporarily.
```
  coverageThreshold: {
    global: {
      branches: 85,
      functions: 95,
      lines: 95,
      statements: 95
    }
  },
  ```

File: apps/entity-service/src/entities/dto/site.dto.ts [hunk 1]
```diff
@@ -7,10 +7,10 @@ import {
   UpdateRequestStatus
 } from "@terramatch-microservices/database/constants/status";
 import { ApiProperty } from "@nestjs/swagger";
-import { Site } from "@terramatch-microservices/database/entities";
+import { Project, Site } from "@terramatch-microservices/database/entities";
 import { FrameworkKey } from "@terramatch-microservices/database/constants/framework";
 import { AdditionalProps, EntityDto } from "./entity.dto";
-
+import { MediaDto } from "./media.dto";
 // TODO: THIS IS A STUB!
```
roguenet, line 14: Please remove this comment.

apps/entity-service/src/entities/dto/site.dto.ts
```diff
@@ -21,7 +21,6 @@ export class SiteLightDto extends EntityDto {
       this.populate(SiteLightDto, {
         ...pickApiProperties(site, SiteLightDto),
         lightResource: true,
-        // these two are untyped and marked optional in the base model.
```
roguenet, line 24: Not a huge deal that this is removed, but it is kind of a useful comment 🤷🏻

apps/entity-service/src/entities/dto/site.dto.ts
```diff
@@ -55,6 +54,12 @@ export class SiteLightDto extends EntityDto {
   @ApiProperty({ nullable: true })
   name: string | null;
 
+  @ApiProperty({
+    nullable: true,
+    description: "The associated project name"
+  })
+  projectName: string | null;
+
```
roguenet, line 62: There are lots of fields on the spreadsheet that aren't represented in this lite DTO, and projectName I didn't think was needed except in the full DTO. Is it actually needed here?

apps/entity-service/src/entities/dto/site.dto.ts
```diff
@@ -85,5 +90,134 @@ export class SiteFullDto extends SiteLightDto {
   lightResource = false;
 
   @ApiProperty()
-  totalSiteReports: number;
+  siteReportsTotal: number;
```
roguenet, line 93: Please review the spreadsheet. In the usage notes for this column I requested that it be renamed to `totalSiteReports` to be consistent with projects. I'm going to skip over the rest of the DTO fields for now; please check what's on the spreadsheet against what you have here and I'll review again when that's done.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,161 @@
...
   readonly FULL_DTO = SiteFullDto;
 
-  async findOne(): Promise<Site> {
-    throw new Error("Method not implemented.");
+  async findOne(uuid: string) {
+    return await Site.findOne({
+      where: { uuid },
+      include: [
+        { association: "framework" },
+        { association: "project", attributes: ["name"], include: [{ association: "organisation" }] }
```
roguenet, line 30: unless projectName / organisationName are needed in the LIght DTO, these associations should be removed. If organisationName _is_ needed, please add the `attributes: ["name"]` bit so the whole org record isn't fetched.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,161 @@
...
-  async findMany(): Promise<PaginatedResult<Site>> {
-    throw new Error("Method not implemented.");
+  async findMany(query: EntityQueryDto, userId?: number, permissions?: string[]): Promise<PaginatedResult<Site>> {
+    const builder = await this.entitiesService.buildQuery(Site, query, [
+      { association: "project", attributes: ["name"] },
+      { association: "framework" }
+    ]);
+
+    if (query.sort != null) {
+      if (["name"].includes(query.sort.field)) {
```
roguenet, line 42: If this only needs to compare against one value, please `===` instead of `array.includes`

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,161 @@
...
+      builder.where({
+        projectId: { [Op.in]: ProjectUser.projectsManageSubquery(userId) }
+      });
+    }
+
+    for (const term of ["name", "projectName"]) {
+      if (query[term] != null) builder.where({ [term]: query[term] });
+    }
+    if (query.search != null) {
+      builder.where({ name: { [Op.like]: `%${query.search}%` } });
```
roguenet, line 70: The existing implementation for sites (see `Site.php` `search` method) matches the name search on sites against site name, project name or org name. I think org name is a bit odd, especially since we don't do that on projects, but I would at least include project name here. Or check with product. Even that feels a bit weird to me.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,161 @@
...
+
+    const hasBeenSubmittedSiteReports = await SiteReport.hasBeenSubmitted()
+      .sites([siteId])
+      .findAll({ attributes: ["id", "siteId", "numTreesRegenerating"] });
+
+    const approvedSiteReports = await SiteReport.approved()
+      .sites([siteId])
+      .findAll({ attributes: ["id", "siteId", "numTreesRegenerating"] });
+
+    const regeneratedTreesCount = sumBy(hasBeenSubmittedSiteReports, "numTreesRegenerating");
```
roguenet, line 92: Another spot where the audit I did in the spreadsheet will help you. This field is not used in the FE. I would rename the `approvedRegeneratedTreesCount` to just `regeneratedTreesCount`, and remove this submitted reports field entirely. 

A general guidance from product that I forgot to put on the tickets is that all aggregated fields on these entities should be based only on approved entities. The fact that some of them in PHP are using the submitted entities instead is a mistake.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,161 @@
...
+      siteReportIds,
+      SiteReport.LARAVEL_TYPE,
+      Demographic.WORKDAYS_TYPE
+    );
+
+    return (
+      (await DemographicEntry.gender().sum("amount", {
+        where: {
+          demographicId: {
+            [Op.or]: [{ [Op.in]: siteReportWorkdays }]
```
roguenet, line 132: When there's only one, there's no reason to wrap it in an `or`. This can be simply
```
demographicId: { [Op.in]: siteReportWorkdays }
```

jest.preset.js
```diff
@@ -5,10 +5,10 @@ module.exports = {
...
   coverageThreshold: {
     global: {
-      branches: 85,
-      functions: 95,
-      lines: 95,
-      statements: 95
+      branches: 50,
+      functions: 50,
+      lines: 50,
+      statements: 50
```
roguenet, line 11: We'll want this back to normal before the final merge.

libs/common/src/lib/policies/site.policy.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    const frameworks = this.permissions
+      .filter(name => name.startsWith("framework-"))
+      .map(name => name.substring("framework-".length) as FrameworkKey);
+    if (frameworks.length > 0) {
+      this.builder.can("read", Site, { frameworkKey: { $in: frameworks } });
+    }
+
+    if (this.permissions.includes("manage-own")) {
```
roguenet, line 18: For `manage-own`, we should also be checking projects in the org, per the current PHP policy. I did this recently for `project-report.policy.ts`, and you can probably just copy that code over.

libs/database/src/lib/entities/site.entity.ts
```diff
@@ -96,6 +108,102 @@ export class Site extends Model<Site> {
   @Column(TEXT)
   descriptionSitingStrategy: string | null;
 
+  @AllowNull
+  @Column(INTEGER.UNSIGNED)
```
roguenet, line 112: According to the current DB schema, this should be a `DECIMAL(15, 1)`. Please carefully audit these fields against the current DB schema to make sure they all match. Once you've done that, I'll take another look over these new fields.

apps/entity-service/src/entities/dto/site.dto.ts
```diff
@@ -54,17 +54,20 @@ export class SiteLightDto extends EntityDto {
...
-  projectName: string | null;
-
   @ApiProperty()
   createdAt: Date;
 
   @ApiProperty()
   updatedAt: Date;
+
+  @ApiProperty({ type: Project })
+  project: Project;
```
roguenet, line 64: Lots of problems here:
* We'd never want to include the full project on another entity's DTO; that's a pattern from PHP we're ditching in v3. Fetching a couple of fields (like UUID and name) is OK, but if an element on the page needs the full project, it can use the project GET connection to pull it.
* A model cannot itself be an ApiProperty - it has to be a DTO
* I still think the project information isn't needed on the light DTO. Did you find a case where that's incorrect? Also applies to `projectUuid` below.

pachonjcl: Agree, we are removing project as a entity from the dto

pachonjcl: Actually the list of sites in admin do require the `projectName` property, so that one will stay on the light dto.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,167 @@
...
+    const frameworkAssociation: Includeable = {
+      association: "framework",
+      attributes: ["name"]
+    };
+    if (query.search != null) {
+      // This is they way that sequelize supports for searching in a joined table
+      projectAssociation.where = { name: { [Op.like]: `%${query.search}%` } };
+      // This is to ensure that the project is not required to be joined (simulating an OR)
+      projectAssociation.required = false;
+    }
```
pachonjcl, line 46: Please take a look to this comments, if you feel this is an anti pattern o there is a better way to handle this.
Happy to change it.

libs/database/src/lib/entities/site-report.entity.ts
```diff
@@ -52,6 +54,10 @@ export class SiteReport extends Model<SiteReport> {
     return chainScope(this, "approved") as typeof SiteReport;
   }
 
+  static hasBeenSubmitted() {
```
roguenet, line 57: This ended up not being needed, I think go ahead and roll back the changes to this file.

libs/database/src/lib/entities/site.entity.ts
```diff
@@ -122,6 +234,10 @@ export class Site extends Model<Site> {
   })
   seedsPlanted: Seeding[] | null;
 
+  get projectName() {
+    return this.project?.name;
+  }
```
roguenet, line 239: Please move this and `projectUuid` getter to both be just below the `project` declaration. It's useful to see at a glance what the dependencies on an association are.

apps/entity-service/src/entities/dto/site.dto.ts
pachonjcl: I think I wrote it, but yes, we need this name on the index endpoint.

apps/entity-service/src/entities/processors/site.processor.ts
pachonjcl: In this case, this doesn't seem possible.
The second block requires the usage of builder and the first block is required to create the builder.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,167 @@
...
 export class SiteProcessor extends EntityProcessor<Site, SiteLightDto, SiteFullDto> {
   readonly LIGHT_DTO = SiteLightDto;
   readonly FULL_DTO = SiteFullDto;
 
-  async findOne(): Promise<Site> {
-    throw new Error("Method not implemented.");
+  async findOne(uuid: string) {
+    return await Site.findOne({
+      where: { uuid },
+      include: [{ association: "framework" }, { association: "project", attributes: ["uuid"] }]
```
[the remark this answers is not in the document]
pachonjcl: name has been added.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,167 @@
...
+
+    const associations = [projectAssociation, frameworkAssociation];
+
+    const builder = await this.entitiesService.buildQuery(Site, query, associations);
+    if (query.sort != null) {
+      if (query.sort.field === "name") {
+        builder.order([query.sort.field, query.sort.direction ?? "ASC"]);
+      } else if (query.sort.field === "projectName") {
+        builder.order(["project", "name", query.sort.direction ?? "ASC"]);
+      } else if (query.sort.field !== "id") {
```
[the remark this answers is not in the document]
pachonjcl: I think you might referring to TM-429 which is another ticket and is for project.
But I think for now no country sorting.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,167 @@
...
+      });
+    }
+
+    return { models: await builder.execute(), paginationTotal: await builder.paginationTotal() };
+  }
+
+  async addFullDto(document: DocumentBuilder, site: Site): Promise<void> {
+    const siteId = site.id;
+
+    const approvedSiteReportsQuery = SiteReport.approvedIdsSubquery({ val: siteId });
```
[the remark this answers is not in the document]
pachonjcl: changed

libs/database/src/lib/entities/site.entity.ts
pachonjcl: moved

jest.preset.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
 
   coverageThreshold: {
     global: {
-      branches: 85,
+      branches: 84,
```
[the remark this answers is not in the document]
pachonjcl: I will do, just before getting merged. I guess need to add more test for this.

apps/entity-service/src/entities/processors/site.processor.ts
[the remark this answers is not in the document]
pachonjcl: changed

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,167 @@
...
+    const seedsPlantedCount = (await Seeding.visible().siteReports(approvedSiteReportsQuery).sum("amount")) ?? 0;
```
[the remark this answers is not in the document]
pachonjcl: ok, implemented the change.

libs/database/src/lib/entities/site-report.entity.ts
pachonjcl: reverted

apps/entity-service/src/entities/dto/site-query.dto.ts
```diff
@@ -0,0 +1,9 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { EntityQueryDto } from "./entity-query.dto";
+import { IsOptional } from "class-validator";
+
+export class SiteQueryDto extends EntityQueryDto {
+  @ApiProperty({ required: false })
+  @IsOptional()
+  projectUuid?: string;
```
roguenet, line 8: Unfortunately, since these entities all share the same endpoint, this will need to get added to the base EntityQueryDto. The way this is implemented now, there won't be any validation at request time because the controller is only checking against the base EntityQueryDto.

apps/entity-service/src/entities/processors/site.processor.ts
roguenet: oh, my mistake

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,167 @@
...
+  }
+
+  async addLightDto(document: DocumentBuilder, site: Site) {
+    document.addData(site.uuid, new SiteLightDto(site));
+  }
+
+  protected async getWorkdayCount(siteId: number, useDemographicsCutoff = false) {
+    const dueAfter = useDemographicsCutoff ? Demographic.DEMOGRAPHIC_COUNT_CUTOFF : undefined;
+
+    const siteReportIds = SiteReport.approvedIdsSubquery({ val: siteId }, { dueAfter });
```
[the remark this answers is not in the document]
roguenet: Reminder to make this change

libs/common/src/lib/policies/site.policy.ts
```diff
@@ -0,0 +1,55 @@
+import { Project, Site, User } from "@terramatch-microservices/database/entities";
+import { FrameworkKey } from "@terramatch-microservices/database/constants/framework";
+import { UserPermissionsPolicy } from "./user-permissions.policy";
+
+export class SitePolicy extends UserPermissionsPolicy {
+  async addRules() {
+    if (this.permissions.includes("sites-read") || this.permissions.includes("view-dashboard")) {
```
roguenet, line 7: There is no `sites-read` permission, please remove.

libs/common/src/lib/policies/site.policy.ts
```diff
@@ -0,0 +1,55 @@
...
+      this.builder.can("read", Site);
+    }
+
+    const frameworks = this.permissions
+      .filter(name => name.startsWith("framework-"))
+      .map(name => name.substring("framework-".length) as FrameworkKey);
+    if (frameworks.length > 0) {
+      this.builder.can("read", Site, { frameworkKey: { $in: frameworks } });
```
roguenet, line 15: This can use `this.frameworks` from the super class instead of building the array itself.

apps/entity-service/src/entities/dto/site-query.dto.ts
pachonjcl: moved to EntityQueryDto

apps/entity-service/src/entities/dto/entity-query.dto.ts
```diff
@@ -4,7 +4,7 @@ import { NumberPage } from "@terramatch-microservices/common/dto/page.dto";
 
 class QuerySort {
   @ApiProperty({ name: "sort[field]", required: false })
-  field: string;
+  field?: string;
```
roguenet, line 7: It's obnoxious, but the query param validation system and the documentation generation system don't always sync up as I'd like. You've made the param not required from a documentation perspective, but I believe you need to add `@IsOptional()` to avoid problems with validation.

apps/entity-service/src/entities/dto/site.dto.ts
```diff
@@ -67,10 +71,13 @@ export class SiteLightDto extends EntityDto {
 }
 
 export type AdditionalSiteFullProps = AdditionalProps<SiteFullDto, SiteLightDto, Omit<Site, "project">>;
+export type AdditionalSiteLightProps = Pick<SiteFullDto, "treesPlantedCount">;
+
+export type AdditionalSiteCombinedProps = AdditionalSiteFullProps & AdditionalSiteLightProps;
```
roguenet, line 76: The full props should always directly extend from the light props because the full props must always be an explicit super set of light props. So I'd do this instead as 
```
export type AdditionalSiteFullProps = AdditionalSiteLightProps & AdditionalProps<SiteFullDto, SiteLightDto, Omit<Site, "project">>;
```
and remove this combined props type.

apps/entity-service/src/entities/dto/entity.dto.ts
```diff
@@ -11,6 +11,8 @@ export type AdditionalProps<FullDto, LightDto, Model> = Pick<
   keyof Omit<FullDto, keyof Model | keyof LightDto>
 >;
 
+export type AdditionalLightProps<LightDto, Model> = Pick<LightDto, keyof Omit<LightDto, keyof Model>>;
```
roguenet, line 14: This was a good addition. However, we can make it a little better. If you add these modifications (and the similar change for `AdditionalSiteFullProps`, then we can reduce to a single utility type definition:
```
diff --git a/apps/entity-service/src/entities/dto/entity.dto.ts b/apps/entity-service/src/entities/dto/entity.dto.ts
index 29207bb..680c2d9 100644
--- a/apps/entity-service/src/entities/dto/entity.dto.ts
+++ b/apps/entity-service/src/entities/dto/entity.dto.ts
@@ -3,13 +3,10 @@ import { JsonApiAttributesInput, pickApiProperties } from "@terramatch-microserv
 import { ApiProperty } from "@nestjs/swagger";
 
 /**
- * A utility type for constructing the "extra props" type of a full DTO based on what's in the
- * full dto, the light dto and the model.
+ * A utility type for constructing the "extra props" type of a DTO based on what's in the dto, the
+ * base type (e.g. the light DTO & base model for a full DTO)
  */
-export type AdditionalProps<FullDto, LightDto, Model> = Pick<
-  FullDto,
-  keyof Omit<FullDto, keyof Model | keyof LightDto>
->;
+export type AdditionalProps<Dto, BaseType> = Pick<Dto, keyof Omit<Dto, keyof BaseType>>;
 
 export abstract class EntityDto {
   protected populate<DTO extends EntityDto>(dtoClass: Type<DTO>, source: JsonApiAttributesInput<DTO>) {
diff --git a/apps/entity-service/src/entities/dto/project.dto.ts b/apps/entity-service/src/entities/dto/project.dto.ts
index ed4f307..16ceb31 100644
--- a/apps/entity-service/src/entities/dto/project.dto.ts
+++ b/apps/entity-service/src/entities/dto/project.dto.ts
@@ -69,7 +69,10 @@ export class ProjectLightDto extends EntityDto {
   updatedAt: Date;
 }
 
-export type AdditionalProjectFullProps = AdditionalProps<ProjectFullDto, ProjectLightDto, Omit<Project, "application">>;
+export type AdditionalProjectFullProps = AdditionalProps<
+  ProjectFullDto,
+  ProjectLightDto & Omit<Project, "application">
+>;
 export type ProjectMedia = Pick<ProjectFullDto, keyof typeof Project.MEDIA>;
 
 export class ANRDto {

```

apps/entity-service/src/entities/dto/entity-query.dto.ts
pachonjcl: added

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,182 @@
...
+      // This is they way that sequelize supports for searching in a joined table
+      projectAssociation.where = { name: { [Op.like]: `%${query.search}%` } };
+      // This is to ensure that the project is not required to be joined (simulating an OR)
+      projectAssociation.required = false;
+    }
+
+    const associations = [projectAssociation, frameworkAssociation];
+
+    const builder = await this.entitiesService.buildQuery(Site, query, associations);
+    if (query.sort != null) {
```
Scriptmatico, line 59: There is too many if statements to make builder instructions, 

Check this fragment:

```
if (query.sort != null) {
  const { field, direction = "ASC" } = query.sort;
  const sortFields = {
    name: [field, direction],
    projectName: ["project", "name", direction],
  };

  if (sortFields[field]) {
    builder.order(sortFields[field]);
  } else if (query.sort.field !== "id") {
    throw new BadRequestException(`Invalid sort field: ${field}`);
  }
}
```


Using a map can create new order instructions without adding more if statements.

apps/entity-service/src/entities/processors/site.processor.ts
```diff
@@ -1,24 +1,182 @@
...
+      builder.where({
+        projectId: { [Op.in]: ProjectUser.projectsManageSubquery(userId) }
+      });
+    }
+
+    for (const term of ["status", "updateRequestStatus", "frameworkKey"]) {
+      if (query[term] != null) builder.where({ [term]: query[term] });
+    }
+
+    if (query.search != null) {
```
Scriptmatico, line 88: just `if(query.search)` makes the work

apps/entity-service/src/entities/processors/site.processor.spec.ts
```diff
@@ -19,21 +35,173 @@ describe("SiteProcessor", () => {
...
+      await expect(processor.findMany({ sort: { field: "status" } } as EntityQueryDto)).rejects.toThrow(
+        BadRequestException
+      );
     });
   });
 
   describe("findOne", () => {
-    it("throws", async () => {
-      await expect(processor.findOne()).rejects.toThrow();
+    it("returns the requested site", async () => {
```
Scriptmatico, line 170: Please follow descriptive naming to specify the expected behavior, like:

`should return a requested site`
 
Sometimes is good to add alternative scenario

`should throw an error when  requested site is not found`


This naming convention is better to make clarity of your test and expected behavior.

apps/entity-service/src/entities/processors/site.processor.spec.ts
```diff
@@ -19,21 +35,173 @@ describe("SiteProcessor", () => {
...
+      const result = await processor.findOne(site.uuid);
+      expect(result.id).toBe(site.id);
     });
   });
 
   describe("DTOs", () => {
-    it("throws", async () => {
-      await expect(processor.addLightDto()).rejects.toThrow();
-      await expect(processor.addFullDto()).rejects.toThrow();
+    it("SiteLightDto is a light resource", async () => {
```
Scriptmatico, line 178: A better name could be: Should serialize a SiteLightDto as a lightweight resource

libs/common/src/lib/policies/site.policy.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    }).compile();
+
+    service = await module.resolve(PolicyService);
+  });
+
+  afterEach(async () => {
+    jest.restoreAllMocks();
+  });
+
+  it("allows reading all sites with view-dashboard permissions", async () => {
```
Scriptmatico, line 28: This test description also can be improved with the naming convention:

`Should allot read all sites with view-dashboard permissions`

pachonjcl: changed

apps/entity-service/src/entities/processors/site.processor.spec.ts
pachonjcl: changed

apps/entity-service/src/entities/processors/site.processor.spec.ts
pachonjcl: changed

apps/entity-service/src/entities/processors/site.processor.ts
pachonjcl: we are following a pattern of comparing with null rather than falsy values.

apps/entity-service/src/entities/processors/site.processor.ts
pachonjcl: The number of sorting parameters will increase as we move forward.
We are going to add the sorting by more complex parameters like tree count and this doesn't seem like a good optimization at this point.

apps/entity-service/src/entities/processors/site.processor.ts
roguenet: Yeah, I agree with Jose on this one. Implicit boolean checks in JS / TS are an anti pattern in my opinion. Much prefer explicit null or value checks.

roguenet (approved):

PR #69, opened 2025-02-27
[TM-1719] Support beneficiaries demographics

roguenet:
https://gfw.atlassian.net/browse/TM-1719

Also rejiggered the demographics endpoint to no longer send a bunch of stubbed collections. That was being done so that the FE would get titles for collections that don't actually exist on this entity so that it can display a title with an empty demographics grid. That collection -> title mapping has now been moved to the FE.


roguenet:
Moving to draft; I just remembered I have some work to do in the unified database service.

egrojMonroy (approved):

LimberHope (approved):

PR #72, opened 2025-02-28
[TM-1759] add policy for sitepolygon

egrojMonroy:


libs/common/src/lib/policies/site-polygon.policy.ts
```diff
@@ -1,10 +1,22 @@
...
 import { UserPermissionsPolicy } from "./user-permissions.policy";
+import { Op } from "sequelize";
 
 export class SitePolygonPolicy extends UserPermissionsPolicy {
   async addRules() {
     if (this.permissions.includes("polygons-manage")) {
-      this.builder.can("manage", SitePolygon);
+      if (this.frameworks.length === 0) {
+        return;
+      }
```
roguenet, line 10: This will break the research service account's access. The only permission it has is `polygons-manage`.

These should be two separate checks:
```
if (this.permissions.includes("polygons-manage")) {
  this.builder.can("manage", SitePolygon);
}

if (this.frameworks.length > 0) {
  ... put site UUID stuff here.
}
```

egrojMonroy: Ok. The change has done, so now the PHP PR will be removed too.

roguenet (approved):

PR #75, opened 2025-03-03
[TM-1759] add siteId filter for site polygons

egrojMonroy:


apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
```diff
@@ -30,6 +30,16 @@ export class SitePolygonQueryDto extends IntersectionType(CursorPage) {
   @IsArray()
   projectId?: string[];
 
+  @ApiProperty({
+    name: "siteId[]",
+    isArray: true,
+    required: false,
+    description: "Filter results by site UUID(s). If specified, the includeTestProjects param is ignored"
```
roguenet, line 37: Please update the docs on `siteId`, `projectId` and `includeTestProjects` to indicate that they're mutually exclusive. Something like instead of the second sentence here, putting `Only one of siteId, projectId and includeTestProjects may be used in a single request`. And then I think throwing a `BadRequestException` in the builder if it sees that more than one of them is present would be a good addition. This is getting complicated enough that it deserves some guardrails.

apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
```diff
@@ -70,7 +72,7 @@ export class SitePolygonQueryDto extends IntersectionType(CursorPage) {
     required: false,
     default: false,
     description:
-      "Include polygons for test projects in the results. If an explicit list of project UUIDs is included in projectId[], this parameter is ignored."
+      "Include polygons for test projects in the results. If an explicit list of project UUIDs is included in projectId[], this parameter is ignored. Only one of siteId, projectId and includeTestProjects may be used in a single request"
```
roguenet, line 75: I was saying to replace the second sentence in each of these, not add on. The new sentence has clearer and more accurate than the old second one.

apps/research-service/src/site-polygons/site-polygons.service.ts
```diff
@@ -54,6 +54,11 @@ export class SitePolygonsService {
     return reportingPeriods;
   }
 
+  async getSiteName(sitePolygon: SitePolygon): Promise<string> {
+    const site = await sitePolygon.loadSite();
+    return site.name;
+  }
```
roguenet, line 60: This is not needed, and is issuing a bunch of extra DB queries. Instead, add the "name" to attributes here: https://github.com/wri/terramatch-microservices/blob/ef8c22f5a4285f3b2e3e594093ae2da6e9a8b542/apps/research-service/src/site-polygons/site-polygon-query.builder.ts#L56 and then you can just do `sitePolygon.site.name`

apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
egrojMonroy: understood. 
now is fixed

apps/research-service/src/site-polygons/site-polygons.service.ts
egrojMonroy: It was added there, but didn't though it was as that simple. Thanks.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -51,6 +51,15 @@ export class SitePolygonsController {
   async findMany(@Query() query: SitePolygonQueryDto): Promise<JsonApiDocument> {
     await this.policyService.authorize("readAll", SitePolygon);
 
+    const { siteId, projectId, includeTestProjects } = query;
```
Scriptmatico, line 54: Despite your code is correct, I see some areas for improvement.

Here you check that just 1 param is sent, this looks correct to me.

In line 75, there is no need of if else statement, due are not mutually exclusive, can't exists more than one query param at the same time:

this could be converted to:

```
if (query.siteId != null) {
  await queryBuilder.filterSiteUuids(query.siteId);
}

if (query.projectId != null) {
  await queryBuilder.filterProjectUuids(query.projectId);
}

// Ensure test projects are excluded only if not included explicitly
if (!query.includeTestProjects) {
  await queryBuilder.excludeTestProjects();
}
```

This is easier to extend without affecting the existing conditions.

egrojMonroy: Cool. Change is done. thanks

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -94,7 +97,7 @@ export class SitePolygonsController {
           indicators,
           establishmentTreeSpecies,
           reportingPeriods,
-          siteName
+          sitePolygon.site?.name
```
roguenet, line 100: :art: You can leave as-is, but I'm pretty sure this `?` isn't needed because site gets an "required" inner join, meaning that a site polygon won't be returned if the site for it isn't found.

roguenet (approved):

roguenet:
I approved, but I see now that there is a unit test failure that needs to be addressed.

roguenet (approved):

Scriptmatico (approved):

PR #77, opened 2025-03-05
[TM-1767] add basic functionality for nursery

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1767)


apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -0,0 +1,122 @@
...
+  async addFullDto(document: DocumentBuilder, nursery: Nursery): Promise<void> {
+    const nuseryId = nursery.id;
+
+    const nurseriesReports = await NurseryReport.nurseries([nuseryId]).findAll({
+      attributes: ["id", "seedlingsYoungTrees"]
+    });
+
+    const seedlingsGrownCount = nurseriesReports.reduce((sum, { seedlingsYoungTrees }) => {
+      return sum + (seedlingsYoungTrees ?? 0);
+    }, 0);
```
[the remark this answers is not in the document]
roguenet: Actually, it looks like `nurseriesReports` is only being used for this and for count. Since I think the report count should include all reports and this value should only include approved reports, I'd probably do them as two isolated queries:
```
const nurseryReportsTotal = await NurseryReport.nurseries([nurseryId]).count();
const seedlingsGrownCount = (await NurseryReport.nurseries([nurseryId]).approved().findAll({ 
  raw: true,
  attributes: [fn('SUM', col('seedlings_young_trees'), 'seedlingsYoungTrees']
})).seedlingsYoungTrees;
```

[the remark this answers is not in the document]
roguenet: You'll need to add the `approved` cope to NurseryReport (can copy from SiteReport)

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -0,0 +1,122 @@
...
+  async addLightDto(document: DocumentBuilder, nursery: Nursery): Promise<void> {
+    const nuseryId = nursery.id;
+
+    const nurseriesReports = await NurseryReport.nurseries([nuseryId]).findAll({
+      attributes: ["id", "seedlingsYoungTrees"]
+    });
+
+    const seedlingsGrownCount = nurseriesReports.reduce((sum, { seedlingsYoungTrees }) => {
+      return sum + (seedlingsYoungTrees ?? 0);
+    }, 0);
```
roguenet, line 114: Same feedback here; use the aggregate query from above instead of fetching all reports.

roguenet (commented):
Looks like there are also several conflicts with the staging branch to sort out.

apps/entity-service/src/entities/dto/nursery.dto.ts
```diff
@@ -0,0 +1,140 @@
...
+  UPDATE_REQUEST_STATUSES,
+  UpdateRequestStatus
+} from "@terramatch-microservices/database/constants/status";
+import { MediaDto } from "./media.dto";
+
+@JsonApiDto({ type: "nurseries" })
+export class NurseryLightDto extends EntityDto {
+  constructor(nursery?: Nursery, props?: AdditionalNurseryLightProps) {
+    super();
+    if (nursery != null) {
```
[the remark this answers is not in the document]
pachonjcl: we are following the same pattern that we have on project.dto.ts, but happy to change if that is the case.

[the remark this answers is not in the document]
roguenet: Oh shoot, I remember why I did that now. It was so the full DTO could call super() without causing a population of the lite properties. I'm a little torn on which is better. This is a little confusing on a drive-by read, but it's a little more efficient. 🤷🏻

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -0,0 +1,122 @@
...
+    } else if (permissions?.includes("projects-manage")) {
+      builder.where({ projectId: { [Op.in]: ProjectUser.projectsManageSubquery(userId) } });
+    }
+
+    for (const term of ["status", "updateRequestStatus", "frameworkKey"]) {
+      if (query[term] != null) builder.where({ [term]: query[term] });
+    }
+
+    if (query.search != null) {
+      builder.where({ name: { [Op.like]: `%${query.search}%` } });
```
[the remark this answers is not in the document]
roguenet: Have you had a chance to follow up with product about this?

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -103,20 +104,22 @@ export class NurseryProcessor extends EntityProcessor<Nursery, NurseryLightDto,
...
+
+    const seedlingsGrownCount =
+      (
+        await NurseryReport.nurseries([nurseryId])
+          .approved()
+          .findAll({
+            raw: true,
+            attributes: [[fn("SUM", col("seedlings_young_trees")), "seedlingsYoungTrees"]]
+          })
+      )[0].seedlingsYoungTrees ?? 0;
```
roguenet, line 117: Since this count is used in both DTOs, please move the code to a private method in the processor to avoid duplication.

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+    await Nursery.truncate();
+
+    const module = await Test.createTestingModule({
+      providers: [{ provide: MediaService, useValue: createMock<MediaService>() }, EntitiesService]
+    }).compile();
+
+    processor = module.get(EntitiesService).createEntityProcessor("nurseries") as NurseryProcessor;
+  });
+
+  describe("findMany", () => {
```
Scriptmatico, line 39: findMany doesnt make sense of the behavior, I recommend to use:

Should return empty array when call findMany with params X
OR
Should return a list of 4 elements when call findMany.

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      const nurseries = await NurseryFactory.createMany(3, { frameworkKey: "hbf" });
+      await NurseryFactory.createMany(3, { frameworkKey: "ppc" });
+      for (const p of await NurseryFactory.createMany(3, { frameworkKey: "terrafund" })) {
+        nurseries.push(p);
+      }
+
+      await expectNurseries(nurseries, {}, { permissions: ["framework-hbf", "framework-terrafund"] });
+    });
+
+    it("searches", async () => {
```
Scriptmatico, line 85: Same here, the naming convention needs to make sense, please use a descriptive test name 

Example:
should return an nursery report when project is correct

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      const n1 = await NurseryFactory.create({ name: "Foo Bar" });
+      const n2 = await NurseryFactory.create({ name: "Baz Foo" });
+      await NurseryFactory.createMany(3);
+
+      await expectNurseries([n1, n2], { search: "foo" });
+    });
+
+    it("filters", async () => {
```
Scriptmatico, line 93: Same here, please use descriptive naming convention, we use to use the should pattern.

```
it("should increase the balance when making a deposit of 100 €", () => {
    const amount = 100;
    const balance = deposit({ amount, bankAccount });
    expect(balance).toBe(1100);
  });
```

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      await expectNurseries([first, second, third], { updateRequestStatus: "awaiting-approval" });
+
+      await expectNurseries([fourth], { projectUuid: p2.uuid });
+    });
+
+    it("should throw an error if the project uuid is not found", async () => {
+      await expect(processor.findMany({ projectUuid: "123" })).rejects.toThrow(BadRequestException);
+    });
+
+    it("sorts by name", async () => {
```
Scriptmatico, line 133: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        { sortField: "name" }
+      );
+      await expectNurseries(
+        [nurseryC, nurseryB, nurseryA],
+        { sort: { field: "name", direction: "DESC" } },
+        { sortField: "name", sortUp: false }
+      );
+    });
+
+    it("sorts by project name", async () => {
```
Scriptmatico, line 150: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        { sortField: "createdAt", sortUp: false }
+      );
+      await expectNurseries(
+        [nurseryB, nurseryC, nurseryA],
+        { sort: { field: "createdAt", direction: "ASC" } },
+        { sortField: "createdAt", sortUp: true }
+      );
+    });
+
+    it("sort by status", async () => {
```
Scriptmatico, line 211: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        { sortField: "startDate", sortUp: false }
+      );
+      await expectNurseries(
+        [nurseryB, nurseryC, nurseryA],
+        { sort: { field: "startDate", direction: "ASC" } },
+        { sortField: "startDate", sortUp: true }
+      );
+    });
+
+    it("sorts by createdAt", async () => {
```
Scriptmatico, line 185: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        { sortField: "projectName" }
+      );
+      await expectNurseries(
+        [nurseryC, nurseryB, nurseryA],
+        { sort: { field: "projectName", direction: "DESC" } },
+        { sortField: "projectName" }
+      );
+    });
+
+    it("sorts by startDate", async () => {
```
Scriptmatico, line 169: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        { sortField: "status" }
+      );
+      await expectNurseries(
+        [nurseryC, nurseryB, nurseryA],
+        { sort: { field: "status", direction: "DESC" } },
+        { sortField: "status", sortUp: false }
+      );
+    });
+
+    it("sort by updateRequestStatus", async () => {
```
Scriptmatico, line 228: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        { sortField: "updateRequestStatus" }
+      );
+      await expectNurseries(
+        [nurseryC, nurseryB, nurseryA],
+        { sort: { field: "updateRequestStatus", direction: "DESC" } },
+        { sortField: "updateRequestStatus", sortUp: false }
+      );
+    });
+
+    it("sorts by organisation name", async () => {
```
Scriptmatico, line 249: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    it("paginates", async () => {
+      const nurseries = sortBy(await NurseryFactory.createMany(25), "id");
+      await expectNurseries(nurseries.slice(0, 10), { page: { size: 10 } }, { total: nurseries.length });
+      await expectNurseries(nurseries.slice(10, 20), { page: { size: 10, number: 2 } }, { total: nurseries.length });
+      await expectNurseries(nurseries.slice(20), { page: { size: 10, number: 3 } }, { total: nurseries.length });
+    });
+  });
+
+  describe("findOne", () => {
```
Scriptmatico, line 299: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+    it("returns the requested nursery", async () => {
```
Scriptmatico, line 300: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    it("should return null when uuid does not exist", async () => {
+      const uuid = "non-existing-uuid";
+      const result = await processor.findOne(uuid);
+      expect(result).toBeNull();
+    });
+  });
+
+  describe("DTOs", () => {
+    it("NurseryLightDto is a light resource", async () => {
```
Scriptmatico, line 314: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      const nursery = await processor.findOne(uuid);
+      const document = buildJsonApi(NurseryLightDto, { forceDataArray: true });
+      await processor.addLightDto(document, nursery);
+      const attributes = document.serialize().data[0].attributes as NurseryLightDto;
+      expect(attributes).toMatchObject({
+        uuid,
+        lightResource: true
+      });
+    });
+    it("includes calculated fields in NurseryFullDto", async () => {
```
Scriptmatico, line 325: Please change it to `should` pattern

apps/entity-service/src/entities/processors/nursery.processor.spec.ts
roguenet: I agree with this feedback for `it()` calls, but for `describe()` I think simply naming the method or behavior that is being described is fine. This block is describing the behavior of the `findMany` method, I think that's pretty clear.

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -60,9 +60,26 @@ export class NurseryProcessor extends EntityProcessor<Nursery, NurseryLightDto,
...
+    // };
+
+    // if (query.search != null) {
+    //   builder.where({
+    //     [Op.and]: [
+    //       { name: { [Op.like]: `%${query.search}%` } }, // Filtra por nombre en el modelo 'Nursery'
+    //       // { '$project.name$': { [Op.like]: `%${query.search}%` } } // Filtra por nombre en el modelo asociado 'Project'
+    //     ]
+    //   });
+    // }
```
roguenet, line 82: Please don't include commented out code.

libs/database/src/lib/util/paginated-query.builder.ts
```diff
@@ -70,6 +70,18 @@ export class PaginatedQueryBuilder<T extends Model<T>> {
     return this;
   }
 
+  // Adds associations to the query, enabling filtering on related model fields.
+  withAssociations(include: Includeable[]) {
```
roguenet, line 74: I like this idea, but for it be more functional, please have it merge into the existing array. Something like this (note that I've added an explicit null check as well:
```
if (include == null || include.length === 0) {
  throw new BadRequestException(...);
}

this.findOptions.include = [...(this.findOptions.include ?? []), ...include];
this.pageTotalFindOptions.include = [...(this.pageTotalFindOptions.include ?? []), ...include];

return this;
```

Note also that I discovered while helping Jorge with the site polygon controller that it's a little more complicated than just adding the includes to the page total find options. If this works for this usecase, then I'm OK with leaving it alone for now, but it will likely need to get more complicated. If you look [here](https://github.com/wri/terramatch-microservices/pull/89/commits/b13cc707c852e38ca4a5bea33c759599303e636c#diff-1127ae2f4b91bef2ef6874a76a9a811d533fa674ecc4a5ab5e16d93a04231f08R78), you'll see that I had to modify the includes, and in fact the only includes that need to be added to the page total find options, are `required` includes, because they're the only ones that modify the output of the `count()` call. Since the only use of this function in this PR doesn't require the include, I think maybe leave off adding it to pageTotalFindOptions for now. I can explore a better generic solution soon.

roguenet: I had an idea about this on a drive this evening, and pulled it together. Also merged the staging branch in so this one is up to date now.

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -61,7 +60,7 @@ export class NurseryProcessor extends EntityProcessor<Nursery, NurseryLightDto,
     }
 
     if (query.search != null) {
-      builder.withAssociations(associations).where({
```
roguenet: Since `withAssociations` doesn't need to do anything special to the page find options anymore, this call was superfluous (the associations are already added above) and I ended up removing the method because it isn't in use.

libs/database/src/lib/entities/nursery-report.entity.ts
```diff
@@ -94,7 +94,7 @@ export class NurseryReport extends Model<NurseryReport> {
   @HasMany(() => TreeSpecies, {
     foreignKey: "speciesableId",
     constraints: false,
-    scope: { speciesableType: NurseryReport.LARAVEL_TYPE, collection: "nursery-seedling" }
+    scope: { speciesable_type: NurseryReport.LARAVEL_TYPE, collection: "nursery-seedling" }
```
roguenet, line 97: I noted this in the commit message, but there was a bug where `count()` wasn't correctly undercasing this column from the scope, even though `findMany` does 🤷🏻 .

roguenet (dismissed):

apps/entity-service/src/entities/processors/nursery.processor.ts
LimberHope: withAssociation is needed because it is used in search, otherwise there are errors with project.name or project.organisation.name

libs/database/src/lib/entities/nursery-report.entity.ts
LimberHope: Yes, this one is weird

apps/entity-service/src/entities/processors/nursery.processor.ts
roguenet: Not with the code I pushed last night. I made changes to how the page count query is generated.

libs/database/src/lib/entities/nursery-report.entity.ts
roguenet: Yeah, it's not the first bug I've found in the underscore system. I'm not sure if it's the base sequelize library at fault or the TS extension, but it's pretty annoying. That's why several of the models have to specify their multi column indicies in the `@Table` decoration, and have this comment:
https://github.com/wri/terramatch-microservices/blob/d6ce98f57876a480390f6c3a2e9d01500cdbde83/libs/database/src/lib/entities/tree-species.entity.ts#L40-L44

apps/entity-service/src/entities/processors/nursery.processor.ts
LimberHope: I just saw it, you're right, it's no longer necessary.

LimberHope: I'll merge it into staging once the release is done

roguenet: Merging the main back into staging is a standard part of the release process. No extra steps will be needed on your end.

roguenet: I don't think there's any reason you can't go ahead and merge the three PRs into the `staging` branches now, if they're ready. That's the whole reason to cut a separate release branch; so that the release process doesn't hold up development progress on other in-flight work.

PR #86, opened 2025-03-13
[TM-1823] add light resource option in researchservice

egrojMonroy:


apps/research-service/src/site-polygons/dto/site-polygon.dto.ts
```diff
@@ -45,25 +45,21 @@ export class ReportingPeriodDto {
...
-      siteId: sitePolygon.siteUuid,
-      geometry,
-      indicators,
-      establishmentTreeSpecies,
-      reportingPeriods,
-      siteName
-    });
+export class SitePolygonLightDto {
+  constructor(sitePolygon: SitePolygon, indicators: IndicatorDto[]) {
+    if (sitePolygon) {
```
roguenet, line 50: Please use explicit null checks (`if (sitePolygon != null)`), and make the params to the constructor optional with `?`

apps/research-service/src/site-polygons/dto/site-polygon.dto.ts
```diff
@@ -117,6 +93,44 @@ export class SitePolygonDto extends JsonApiAttributes<SitePolygonDto> {
...
+}
+
+export class SitePolygonFullDto extends SitePolygonLightDto {
+  constructor(
+    sitePolygon: SitePolygon,
+    indicators: IndicatorDto[],
+    establishmentTreeSpecies: TreeSpeciesDto[],
+    reportingPeriods: ReportingPeriodDto[]
+  ) {
+    super(sitePolygon, indicators);
```
roguenet, line 107: As with the entity DTOs, we want to skip passing the params to the super class because it causes multiple Object.assign() calls.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -46,13 +46,14 @@ export class SitePolygonsController {
...
   @ApiOperation({ operationId: "sitePolygonsIndex", summary: "Get all site polygons" })
-  @JsonApiResponse({ data: SitePolygonDto, pagination: "cursor" })
+  @JsonApiResponse({ data: SitePolygonFullDto, pagination: "cursor" })
   @ExceptionResponse(UnauthorizedException, { description: "Authentication failed." })
   @ExceptionResponse(BadRequestException, { description: "One or more query param values is invalid." })
   async findMany(@Query() query: SitePolygonQueryDto): Promise<JsonApiDocument> {
     await this.policyService.authorize("readAll", SitePolygon);
 
-    const { siteId, projectId, includeTestProjects, missingIndicator, presentIndicator } = query;
+    const { siteId, projectId, includeTestProjects, missingIndicator, presentIndicator, isLightData } = query;
```
roguenet, line 55: Hmm, I think this approach is kind of awkward because we're documenting the endpoint as using the full DTO, but we sometimes return the light DTO instead, which means the typing in the FE is going to be expecting properties to be defined that aren't. 

I think let's update the JsonApiResponse decorator above to show that either one can be returned with:
```
@JsonApiResponse([{ data: SitePolygonFullDto, pagination: "cursor" }, { data: SitePolygonLightDto, pagination: "number" }])
```
That also has the nice side effect of showing that when it's the light DTO, it's using number pagination. It might even be worth adding a `BadRequestException` if the user requests light DTOs but has cursor pagination indicated. 

I'm starting to feel a little weird about how the overloading of this controller is going; starting to feel pretty fragile. We can keep it for now, but if we see more differences needed between what the FE needs and what the research team needs, I think we'll want to extract the FE needs into its own controller.

apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
```diff
@@ -98,4 +98,11 @@ export class SitePolygonQueryDto extends IntersectionType(CursorPage, NumberPage
   })
   @IsOptional()
   page?: CursorPage | NumberPage;
+
+  @ApiProperty({
+    required: false,
+    default: false,
+    description: "Wheter to include the complete sitePolygon Dto or not"
+  })
+  isLightData?: boolean;
```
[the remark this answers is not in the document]
egrojMonroy: changed the property name and also created a base.dto

apps/research-service/src/site-polygons/dto/site-polygon.dto.ts
egrojMonroy: ok

apps/research-service/src/site-polygons/site-polygons.controller.ts
egrojMonroy: Now both dtos are expected in the jsonApiReponse.
Also I made the change for pagination and badrequest exception.

roguenet:
I think you need to push your branch - github isn't showing any new changes since my last review.

apps/research-service/src/site-polygons/dto/site-polygon.dto.ts
egrojMonroy: removed the params in super for full sto.

libs/common/src/lib/dto/base.dto.ts
```diff
@@ -0,0 +1,18 @@
+import { ApiProperty } from "@nestjs/swagger";
+import { Type } from "@nestjs/common";
+
+export abstract class BaseDto {
+  /**
+   * All DTOs must include a "lightResource" boolean to indicate if the resource that the
+   * client currently has cached is a full version of that resource or not.
+   **/
```
roguenet, line 8: Hmm, I don't love this name or this comment. It's not true that all DTOs need this, it's just DTOs that have a light / full combination. Maybe something like `HybridSupportDto` or `PartialDto`

apps/research-service/src/site-polygons/dto/site-polygon.dto.ts
```diff
@@ -43,27 +43,20 @@ export class ReportingPeriodDto {
...
-      ...pickApiProperties(sitePolygon, SitePolygonDto),
-      name: sitePolygon.polyName,
-      siteId: sitePolygon.siteUuid,
-      geometry,
-      indicators,
-      establishmentTreeSpecies,
-      reportingPeriods,
-      siteName
-    });
+@JsonApiDto({ type: "sitePolygons" })
```
roguenet, line 47: Looks like this decorator is duplicated now.

apps/research-service/src/site-polygons/site-polygons.service.ts
```diff
@@ -95,4 +101,15 @@ export class SitePolygonsService {
       throw e;
     }
   }
+  async buildLightDto(sitePolygon: SitePolygon): Promise<SitePolygonLightDto> {
```
roguenet, line 104: Please include an empty line between method definitions.

roguenet (commented):
Looks like the tests need more coverage.

apps/research-service/src/site-polygons/dto/site-polygon.dto.ts
egrojMonroy: removed

apps/research-service/src/site-polygons/site-polygons.service.ts
egrojMonroy: added

libs/common/src/lib/dto/base.dto.ts
egrojMonroy: changed the name and the comment

libs/common/src/lib/dto/hybridSupport.dto.ts
```diff
@@ -1,11 +1,11 @@
 import { ApiProperty } from "@nestjs/swagger";
 import { Type } from "@nestjs/common";
 
-export abstract class BaseDto {
+export abstract class HybridSupportDto {
```
roguenet, line 4: Our file names in v3 use kebab case, so this file should be named `hybrid-support.dto.ts`

egrojMonroy: changed

roguenet (approved):

PR #89, opened 2025-03-17
[TM-1826] add search to query by poly_name and site.name

egrojMonroy:


apps/research-service/src/site-polygons/site-polygon-query.builder.ts
```diff
@@ -90,6 +90,21 @@ export class SitePolygonQueryBuilder extends PaginatedQueryBuilder<SitePolygon>
...
+  async addSearch(searchTerm: string) {
+    if (!searchTerm) return this;
+
+    return this.where(
+      Sequelize.or(
+        Sequelize.where(Sequelize.fn("LOWER", Sequelize.col("site.name")), {
+          [Op.like]: `%${searchTerm.toLowerCase()}%`
+        }),
+        Sequelize.where(Sequelize.fn("LOWER", Sequelize.col("poly_name")), {
+          [Op.like]: `%${searchTerm.toLowerCase()}%`
```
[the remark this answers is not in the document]
cesarLima1: Done.
We didn't receive any specific instructions on whether the field should match exactly, so I switched to the fuzzy search approach.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -99,6 +98,9 @@ export class SitePolygonsController {
     if (!query.includeTestProjects && query.siteId == null && query.projectId == null) {
       await queryBuilder.excludeTestProjects();
     }
+    if (query.search) {
```
roguenet, line 101: Please use explicit null checks (`if (query.search != null)`)

apps/research-service/src/site-polygons/site-polygon-query.builder.ts
```diff
@@ -91,6 +103,14 @@ export class SitePolygonQueryBuilder extends PaginatedQueryBuilder<SitePolygon>
     return this.where({ projectId: { [Op.in]: filterProjects.map(({ id }) => id) } }, this.siteJoin);
   }
 
+  async addSearch(searchTerm: string) {
+    if (!searchTerm) return this;
```
roguenet, line 107: I think this is superfluous. Typescript should be ensuring that `searchTerm` is not null (not totally safe because something could sneak in at runtime, but if the function doesn't declare that it accepts undefined or null, it should be considered OK to skip the null check), and the place where this function is called is checking it.

apps/research-service/src/site-polygons/site-polygon-query.builder.ts
```diff
@@ -91,6 +103,14 @@ export class SitePolygonQueryBuilder extends PaginatedQueryBuilder<SitePolygon>
...
+
+    return this.where({
+      [Op.or]: [{ "$site.name$": { [Op.like]: `%${searchTerm}%` } }, { polyName: { [Op.like]: `%${searchTerm}%` } }]
```
roguenet, line 110: In the comments you mention that you've switched to the fuzzy search, but this doesn't look like it's using it.

roguenet (commented):
Looks like some additional test are required to get coverage up to the required levels.

apps/research-service/src/site-polygons/site-polygons.controller.ts
egrojMonroy: done

apps/research-service/src/site-polygons/site-polygon-query.builder.ts
egrojMonroy: remove.

apps/research-service/src/site-polygons/site-polygon-query.builder.ts
egrojMonroy: Now Is like tree.service. 
Nonetheless now I have doubt about what is fuzzy search. 
thought this was a fuzzy search too.

apps/research-service/src/site-polygons/site-polygon-query.builder.ts
```diff
@@ -104,10 +104,13 @@ export class SitePolygonQueryBuilder extends PaginatedQueryBuilder<SitePolygon>
...
   async addSearch(searchTerm: string) {
-    if (!searchTerm) return this;
-
     return this.where({
-      [Op.or]: [{ "$site.name$": { [Op.like]: `%${searchTerm}%` } }, { polyName: { [Op.like]: `%${searchTerm}%` } }]
+      [Op.or]: [
+        { "$site.name$": { [Op.like]: `${searchTerm}%` } },
+        { polyName: { [Op.like]: `% ${searchTerm}%` } },
+        { "$site.name$": { [Op.like]: `${searchTerm}%` } },
+        { polyName: { [Op.like]: `% ${searchTerm}%` } }
```
roguenet, line 112: The two `polyName` queries are the same, and the two site name queries are the same. I think you're looking for this: 
```
        { "$site.name$": { [Op.like]: `${searchTerm}%` } },
        { "$site.name$": { [Op.like]: `% ${searchTerm}%` } },
        { polyName: { [Op.like]: `${searchTerm}%` } },
        { polyName: { [Op.like]: `% ${searchTerm}%` } }
 ```

File: apps/research-service/src/site-polygons/dto/site-polygon.dto.ts [hunk 6]
```diff
@@ -117,6 +89,52 @@ export class SitePolygonDto extends JsonApiAttributes<SitePolygonDto> {
   })
   indicators: IndicatorDto[];
 
+  @ApiProperty({ description: "The name of the associated Site." })
+  siteName: string;
+}
+export class SitePolygonFullDto extends SitePolygonLightDto {
```
roguenet, line 95: Please make sure there are newlines between classes and methods.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -98,25 +107,23 @@ export class SitePolygonsController {
...
-      );
+      if (lightResource) {
+        document.addData(sitePolygon.uuid, new SitePolygonLightDto(sitePolygon, indicators));
+      } else {
+        const establishmentTreeSpecies = await this.sitePolygonService.getEstablishmentTreeSpecies(sitePolygon);
+        const reportingPeriods = await this.sitePolygonService.getReportingPeriods(sitePolygon);
+
+        document.addData(
+          sitePolygon.uuid,
+          new SitePolygonFullDto(sitePolygon, indicators, establishmentTreeSpecies, reportingPeriods)
```
roguenet, line 124: It seems like this controller should be using the `buildLightDto` / `buildFullDto` methods from the service instead of creating the DTOs itself?

apps/research-service/src/site-polygons/site-polygon-query.builder.ts
roguenet: searching with like `'%${searchTerm}%'` will only  match against strings that are exactly the same as the search term, ignoring case. "Fuzzy" search just means that the search term can be matched against multiple words in a phrase. A truly fuzzy search would do some work to do partial substring matching and such, which this approach doesn't do, but that requires a lot more infrastructure than a simple like query, so this is a nice middle ground. Matching against `'% ${seachTerm}'` and ``${searchTerm}%'` gets us matches where the search term is at the start of a word, but not in the middle of a word.

roguenet (commented):
I think the code looks good, but there's a spec failure in the most recent push.

apps/research-service/src/site-polygons/site-polygons.controller.spec.ts
```diff
@@ -57,6 +60,22 @@ describe("SitePolygonsController", () => {
       await expect(controller.findMany({})).rejects.toThrow(UnauthorizedException);
     });
 
+    it("should throw an error if the countrselectedparams error is thrown", async () => {
```
Scriptmatico, line 63: Both tests has the same name : `should throw an error if the countrselectedparams error is thrown`

Also, the message description is not clear, I can't understand the behavior with the description. Can you please be more explicit?

apps/research-service/src/site-polygons/site-polygons.controller.spec.ts
```diff
@@ -129,6 +148,15 @@ describe("SitePolygonsController", () => {
       expect(builder.filterProjectUuids).not.toHaveBeenCalled();
       expect(builder.excludeTestProjects).toHaveBeenCalled();
     });
+
+    it("will honor siteId", async () => {
```
Scriptmatico, line 152: You are missing the description convention here ("Should .... honor.. siteid... when....")

apps/research-service/src/site-polygons/site-polygons.controller.spec.ts
egrojMonroy: Yes indeed, changed some of the names to the "should" pattern.

apps/research-service/src/site-polygons/site-polygons.controller.spec.ts
egrojMonroy: done

egrojMonroy: done

roguenet (approved):

PR #94, opened 2025-03-19
[TM-1826] add light resource attrib

egrojMonroy:
Return lightResource attribute for lightResource data

### DIFF
--- a/apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
+++ b/apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
@@ -88,7 +88,8 @@ export class SitePolygonQueryDto extends IntersectionType(CursorPage, NumberPage
   })
   @IsBoolean()
   @Transform(({ value }) => (value === "true" ? true : value === "false" ? false : undefined))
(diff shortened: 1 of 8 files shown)


roguenet (approved):

apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts
```diff
@@ -86,7 +86,10 @@ export class SitePolygonQueryDto extends IntersectionType(CursorPage, NumberPage
     description:
       "Include polygons for test projects in the results. Only one of siteId, projectId and includeTestProjects may be used in a single request"
   })
-  includeTestProjects?: boolean;
+  @IsBoolean()
+  @Transform(({ value }) => (value === "true" ? true : value === "false" ? false : undefined))
+  @IsOptional()
```
roguenet, line 91: This IsOptional is not needed. In fact, I discovered today that's not doing anything on any of these properties, but I'm going to look into a better solution soon and not worry about it for this PR. For now, just leave it off these two booleans.

roguenet (approved):

PR #95, opened 2025-03-20
Feat/[TM-1768] entity get index project report

LimberHope:
Task Link


apps/entity-service/src/entities/dto/project-report.dto.ts
```diff
@@ -0,0 +1,304 @@
...
+  workdaysPaid: number | null;
+
+  @ApiProperty()
+  createdAt: Date;
+
+  @ApiProperty()
+  updatedAt: Date;
+}
+
+// export type AdditionalProjectReportLightProps = Pick<ProjectReportLightDto, "treesPlantedCount">;
```
roguenet, line 76: Please make sure not to leave commented out code in place. (here and two lines down)

apps/entity-service/src/entities/processors/project-report.processor.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+    ) {
+      const { models, paginationTotal } = await processor.findMany(query as EntityQueryDto, userId, permissions);
+      expect(models.length).toBe(expected.length);
+      expect(paginationTotal).toBe(total);
+
+      const sorted = sortBy(expected, sortField);
+      if (!sortUp) reverse(sorted);
+      expect(models.map(({ id }) => id)).toEqual(sorted.map(({ id }) => id));
+    }
+    it("should returns project reports", async () => {
```
roguenet, line 58: Please leave an empty line between test blocks (same as classes and methods)

apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -0,0 +1,195 @@
...
+    for (const term of [
+      "status",
+      "updateRequestStatus",
+      "frameworkKey",
+      "organisationUuid",
+      "country",
+      "projectUuid"
+    ]) {
+      if (query[term] != null) {
+        const field = associationFieldMap[term] || term;
```
roguenet, line 96: Please use `??` for defaults

libs/database/src/lib/entities/site-report.entity.ts
```diff
@@ -31,7 +31,8 @@ type ApprovedIdsSubqueryOptions = {
   incomplete: { where: { status: { [Op.notIn]: COMPLETE_REPORT_STATUSES } } },
   sites: (ids: number[] | Literal) => ({ where: { siteId: { [Op.in]: ids } } }),
   approved: { where: { status: { [Op.in]: SiteReport.APPROVED_STATUSES } } },
-  dueBefore: (date: Date | string) => ({ where: { dueAt: { [Op.lt]: date } } })
+  dueBefore: (date: Date | string) => ({ where: { dueAt: { [Op.lt]: date } } }),
+  reportsTask: (taskId: number) => ({ where: { taskId: taskId } })
```
roguenet, line 35: I think this would be more consistent if simply named `task`. For instance, that would be in line with the `sites` scope above. This scope is attached to a report, so including the word `report` in the scope is redundant.

libs/database/src/lib/entities/nursery-report.entity.ts
```diff
@@ -24,7 +24,8 @@ import { Subquery } from "../util/subquery.builder";
 @Scopes(() => ({
   incomplete: { where: { status: { [Op.notIn]: COMPLETE_REPORT_STATUSES } } },
   nurseries: (ids: number[] | Literal) => ({ where: { nurseryId: { [Op.in]: ids } } }),
-  approved: { where: { status: { [Op.in]: NurseryReport.APPROVED_STATUSES } } }
+  approved: { where: { status: { [Op.in]: NurseryReport.APPROVED_STATUSES } } },
+  reportsTask: (taskId: number) => ({ where: { taskId: taskId } })
```
roguenet, line 28: Same feedback here as on site report - I think this would be better named simply `task`

apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -0,0 +1,195 @@
...
+    const seedsPlantedCount = (await Seeding.visible().siteReports(siteReportsIdsTaks).sum("amount")) ?? 0;
+    const treesPlantedCount =
+      (await TreeSpecies.visible().collection("tree-planted").siteReports(siteReportsIdsTaks).sum("amount")) ?? 0;
+    const approvedSiteReportsFromTask = await SiteReport.approved()
+      .reportsTask(taskId)
+      .findAll({ attributes: ["id", "siteId", "numTreesRegenerating"] });
+    const regeneratedTreesCount = sumBy(approvedSiteReportsFromTask, "numTreesRegenerating");
+    const siteReportsCount = await SiteReport.reportsTask(taskId).count();
+    const nurseryReportsCount = await NurseryReport.reportsTask(taskId).count();
+    const migrated = !!projectReport.oldModel;
```
roguenet, line 135: I'm generally not a fan of the `!!` construction. There are lots of places in JS where folks have depended on implicit boolean conversion in the past, and while it's safer with modern tools, it still gives me the creeps a bit. This is the same as cases where I prefer `thing == null` over `!thing` - this is essentially a null check, let's treat it that way explicitly: `projectReport.oldModel != null`

apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -0,0 +1,195 @@
...
+    const seedlingsGrown = await this.getSeedlingsGrown(projectReport);
+    const siteReportsUnsubmittedIdsTask = await ProjectReport.siteReportsUnsubmittedIdsTaskSubquery([taskId]);
+    const nonTreeTotal = (await Seeding.visible().siteReports(siteReportsUnsubmittedIdsTask).sum("amount")) ?? 0;
+    const readableCompletionStatus = await this.getReadableCompletionStatus(projectReport.completion);
+    const createdByUser = projectReport.user ? projectReport.user : null;
```
roguenet, line 140: This ternary isn't doing much - there's no reason this can't be simply `const createdByUser = projectReport.user`. If you need explicitly null over undefined, use `?? null` instead.

apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -0,0 +1,195 @@
...
+    };
+    document.addData(projectReport.uuid, new ProjectReportFullDto(projectReport, props));
+  }
+
+  async addLightDto(document: DocumentBuilder, projectReport: ProjectReport) {
+    document.addData(projectReport.uuid, new ProjectReportLightDto(projectReport));
+  }
+
+  protected async getReportTitle(projectReport: ProjectReport) {
+    if (projectReport.frameworkKey == "ppc" || !projectReport.dueAt) {
```
roguenet, line 166: Please use explicit null checks `projectReport.dueAt == null`

apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -0,0 +1,195 @@
...
+      return projectReport.title ?? "";
+    } else {
+      const dueAt = new Date(projectReport.dueAt);
+      dueAt.setMonth(dueAt.getMonth() - 1);
+      const wEnd = dueAt.toLocaleString("en-US", { month: "long", year: "numeric" });
+
+      dueAt.setMonth(dueAt.getMonth() - 5);
+      const wStart = dueAt.toLocaleString("en-US", { month: "long" });
```
roguenet, line 174: Is `en-US` what we're using for date formatting on the PHP side? I have a faint memory that in the FE, we typically prefer `en-GB` for date formatting when the user's locale isn't easily available (although we could fetch it here, and maybe that would be better?), but I'm not 100% sure about that.

apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      return `Project Report  for ${wStart} - ${wEnd}`;
```
roguenet, line 176: There's an extra space after `Report`

libs/database/src/lib/entities/project-report.entity.ts
```diff
@@ -63,6 +75,27 @@ export class ProjectReport extends Model<ProjectReport> {
     return builder.literal;
   }
 
+  static siteReportIdsTaksSubquery(taskIds: number[], opts: ApprovedIdsSubqueryOptions = {}) {
```
roguenet, line 78: `Taks` should be `Task`

roguenet (commented):
A few stylistic comments, but this is looking really close!

roguenet (commented):
This looks good to me, but it looks like the specs need some attention.

apps/unified-database-service/src/airtable/entities/airtable-entity.spec.ts
```diff
@@ -607,13 +609,14 @@ describe("AirtableEntity", () => {
       ).reduce((total, { amount }) => total + amount, 0);
       // make sure hidden is ignored
       await TreeSpeciesFactory.forProjectReportNurserySeedling.create({ speciesableId: ppcReport.id, hidden: true });
+      await TaskFactory.create();
 
       // TODO this might start causing problems when Task is implemented in this codebase and we have a factory
       // that's generating real records
```
roguenet, line 615: Please also remove this comment now that it's no longer relevant.

roguenet (dismissed):

roguenet (approved):

PR #101, opened 2025-03-21
Feat/TM-1770 entity get index nursery report

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1770)


apps/entity-service/src/entities/dto/nursery-report.dto.ts
```diff
@@ -0,0 +1,237 @@
...
+  organisationUuid: string | null;
+
+  @ApiProperty()
+  updatedAt: Date;
+
+  @ApiProperty({ nullable: true })
+  submittedAt: Date | null;
+
+  @ApiProperty({ nullable: true })
+  taskId: number | null;
```
roguenet, line 80: I believe we want to be using `taskUuid` here. See the project report entity / dto.

apps/entity-service/src/entities/dto/nursery-report.dto.ts
```diff
@@ -0,0 +1,237 @@
...
+  projectName: string | null;
+
+  @ApiProperty({
+    nullable: true,
+    description: "The associated project uuid"
+  })
+  projectUuid: string | null;
+
+  @ApiProperty({ nullable: true })
+  taskId: number | null;
```
roguenet, line 206: Do we need both `taskId` and `taskUuid` for some reason? I wouldn't expect the FE to care about the integer ID.

apps/entity-service/src/entities/processors/nursery-report.processor.spec.ts
```diff
@@ -0,0 +1,409 @@
...
+      const p1 = await ProjectFactory.create({ country: "MX", organisationId: org1.id });
+      const p2 = await ProjectFactory.create({ country: "CA", organisationId: org2.id });
+      await ProjectUserFactory.create({ userId, projectId: p1.id });
+      await ProjectUserFactory.create({ userId, projectId: p2.id });
+      const n1 = await NurseryFactory.create({ projectId: p1.id });
+      const n2 = await NurseryFactory.create({ projectId: p2.id });
+      // n1.project = await n1.$get("project");
+      // n2.project = await n2.$get("project");
+      // n1.project.organisation = await n1.project.$get("organisation");
+      // n2.project.organisation = await n2.project.$get("organisation");
```
roguenet, line 123: Please don't include commented out code.

apps/entity-service/src/entities/dto/nursery-report.dto.ts
```diff
@@ -0,0 +1,237 @@
...
+  feedback: string | null;
+
+  @ApiProperty({ nullable: true })
+  feedbackFields: string[] | null;
+
+  @ApiProperty()
+  nothingToReport: boolean;
+
+  @ApiProperty()
+  readableCompletionStatus: string;
```
roguenet, line 164: This should not be getting calculated in the BE. The FE has everything it needs to map `status` to a readable string.

apps/entity-service/src/entities/dto/nursery-report.dto.ts
```diff
@@ -0,0 +1,237 @@
...
+export class NurseryReportLightDto extends EntityDto {
+  constructor(nurseryReport?: NurseryReport, props?: AdditionalNurseryReportLightProps) {
+    super();
+    if (nurseryReport != null) {
+      this.populate(NurseryReportLightDto, {
+        ...pickApiProperties(nurseryReport, NurseryReportLightDto),
+        lightResource: true,
+        // these two are untyped and marked optional in the base model.
+        createdAt: nurseryReport.createdAt as Date,
+        updatedAt: nurseryReport.createdAt as Date,
```
Scriptmatico, line 19: It is a typo here? the createdAt is being assigned to both properties.

Additional to this, The casting as Date only works if nurseryReport.createdAt and nurseryReport.updatedAt are actually of type Date or can be properly converted to Date. If they are stored as strings (e.g., ISO date strings), this casting will not work properly.

you can use>

```
createdAt: new Date(nurseryReport.createdAt),
updatedAt: new Date(nurseryReport.updatedAt),
```

apps/entity-service/src/entities/dto/nursery-report.dto.ts
```diff
@@ -0,0 +1,237 @@
...
+export class NurseryReportFullDto extends NurseryReportLightDto {
+  constructor(nurseryReport: NurseryReport, props?: AdditionalNurseryReportFullProps) {
+    super();
+    if (nurseryReport != null) {
+      this.populate(NurseryReportFullDto, {
+        ...pickApiProperties(nurseryReport, NurseryReportFullDto),
+        lightResource: false,
+        // these two are untyped and marked optional in the base model.
+        createdAt: nurseryReport.createdAt as Date,
+        updatedAt: nurseryReport.createdAt as Date,
```
Scriptmatico, line 109: Check comment in line 17, I think will be the same case that my previous comment.

apps/entity-service/src/entities/processors/nursery-report.processor.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    const associationFieldMap = {
+      nurseryUuid: "$nursery.uuid$",
+      organisationUuid: "$nursery.project.organisation.uuid$",
+      country: "$nursery.project.country$",
+      projectUuid: "$nursery.project.uuid$"
+    };
+
+    for (const term of [
```
Scriptmatico, line 93: you can improve readability here by moving the terms to>

```
// This to the top of the declaration
const termsToFilter = [
  "status",
  "updateRequestStatus",
  "frameworkKey",
  "nurseryUuid",
  "organisationUuid",
  "country",
  "projectUuid"
];
```

and

```
termsToFilter.forEach(term => {
  const field = associationFieldMap[term] || term;
  if (query[term] != null) {
    builder.where({ [field]: query[term] });
  }
});
```

apps/entity-service/src/entities/processors/nursery-report.processor.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  async addFullDto(document: DocumentBuilder, nurseryReport: NurseryReport): Promise<void> {
+    const nurseryReportId = nurseryReport.id;
+    const reportTitle = await this.getReportTitle(nurseryReport);
+    const projectReportTitle = await this.getProjectReportTitle(nurseryReport);
+    const readableCompletionStatus = await this.getReadableCompletionStatus(nurseryReport.completion);
+    const createdByUser = await User.findOne({ where: { id: nurseryReport?.createdBy } });
+    const approvedByUser = await User.findOne({ where: { id: nurseryReport?.approvedBy } });
+    const migrated = nurseryReport.oldModel != null;
+    const props: AdditionalNurseryReportFullProps = {
```
Scriptmatico, line 136: A  little of refactor here for using parallel responses.

```
const props: AdditionalNurseryReportFullProps = {
    reportTitle,
    projectReportTitle,
    readableCompletionStatus,
    createdByUser,
    approvedByUser,
    migrated: nurseryReport.oldModel != null,
    ...(this.entitiesService.mapMediaCollection(
      mediaCollection,
      NurseryReport.MEDIA
    ) as NurseryReportMedia)
  };
```

apps/entity-service/src/entities/dto/nursery-report.dto.ts
roguenet: That's a good catch including `createdAt` twice. The casting however is fine. The only reason we have to include them here at all is because, as the comment states, they're marked as optional in the base Model class because not all models have timestamps, but they are in fact always included, and are converted to Dates in these entities models.

apps/entity-service/src/entities/processors/nursery-report.processor.ts
```diff
@@ -0,0 +1,183 @@
...
+    return { models: await builder.execute(), paginationTotal: await builder.paginationTotal() };
+  }
+
+  async addFullDto(document: DocumentBuilder, nurseryReport: NurseryReport): Promise<void> {
+    const nurseryReportId = nurseryReport.id;
+    const reportTitle = await this.getReportTitle(nurseryReport);
+    const projectReportTitle = await this.getProjectReportTitle(nurseryReport);
+    const readableCompletionStatus = await this.getReadableCompletionStatus(nurseryReport.completion);
+    const createdByUser = await User.findOne({ where: { id: nurseryReport?.createdBy } });
+    const approvedByUser = await User.findOne({ where: { id: nurseryReport?.approvedBy } });
```
roguenet, line 134: These should be getting included in the initial fetch with a join; no reason to pull them in separate queries after the fact.

apps/entity-service/src/entities/dto/nursery-report.dto.ts
```diff
@@ -105,8 +105,8 @@ export class NurseryReportFullDto extends NurseryReportLightDto {
         ...pickApiProperties(nurseryReport, NurseryReportFullDto),
         lightResource: false,
         // these two are untyped and marked optional in the base model.
-        createdAt: nurseryReport.createdAt as Date,
-        updatedAt: nurseryReport.createdAt as Date,
+        createdAt: new Date(nurseryReport.createdAt),
+        updatedAt: new Date(nurseryReport.updatedAt),
```
roguenet, line 109: `new Date()` is not necessary; please go back to `as Date`

roguenet (dismissed):

roguenet (dismissed):

roguenet (approved):

PR #102, opened 2025-03-21
[TM-1781] NestJS REPL for services access in local and remote environments.

roguenet:
https://gfw.atlassian.net/browse/TM-1781

See the changes to the README for some relevant notes.


pachonjcl (approved):
If I can add a comment.
I can't see where my cursor is and need to guess with arrows otherwise looks good.

roguenet:
> If I can add a comment. I can't see where my cursor is and need to guess with arrows otherwise looks good.

Yeah, I'll see if I can do something about that. Was that happening for you locally, remotely, or both? I didn't mention it above, but in case you didn't go digging in the deploy logs to find it, you can test the remote execution with `entity-service` and `test` environment.

roguenet:
> > If I can add a comment. I can't see where my cursor is and need to guess with arrows otherwise looks good.
> 
> Yeah, I'll see if I can do something about that. Was that happening for you locally, remotely, or both? I didn't mention it above, but in case you didn't go digging in the deploy logs to find it, you can test the remote execution with `entity-service` and `test` environment.

Did a little experimenting and found that for me, the cursor shows in remote execution, but not locally. This appears to be a problem with how it's launching through `NX`. I'll find a workaround before I merge this PR.

pachonjcl (approved):

PR #103, opened 2025-03-24
[TM-1769] add fields and basic structure for site report

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1769)


apps/entity-service/src/entities/dto/site-report.dto.ts
```diff
@@ -0,0 +1,302 @@
...
+  organisationUuid: string | null;
+
+  @ApiProperty()
+  updatedAt: Date;
+
+  @ApiProperty({ nullable: true })
+  submittedAt: Date | null;
+
+  @ApiProperty({ nullable: true })
+  taskId: number | null;
```
roguenet, line 81: I think we want to be sending task UUIDs to the client instead of the integer ID.

apps/entity-service/src/entities/dto/site-report.dto.ts
```diff
@@ -0,0 +1,302 @@
...
+  createdBy: number | null;
+
+  @ApiProperty({ nullable: true })
+  createdByUser: User | null;
+
+  @ApiProperty({ nullable: true })
+  approvedBy: number | null;
+
+  @ApiProperty({ nullable: true })
+  approvedByUser: User | null;
```
roguenet, line 178: Including a bare User record in a DTO is an anti pattern. We should never be asking the JSON API serialization code to encode a full DB model. This should be a light weight DTO including just what's needed. Probably just the name fields? Also, are the `createdBy` and `approvedBy` numerical user IDs actually useful for the client? I wouldn't expect it to want them.

apps/entity-service/src/entities/dto/site-report.dto.ts
```diff
@@ -0,0 +1,302 @@
...
+  feedback: string | null;
+
+  @ApiProperty({ nullable: true })
+  feedbackFields: string[] | null;
+
+  @ApiProperty()
+  nothingToReport: boolean;
+
+  @ApiProperty()
+  readableCompletionStatus: string;
```
roguenet, line 160: The FE has everything it needs to generate a readable string from `status`; please don't calculate this in the BE.

apps/entity-service/src/entities/dto/site-report.dto.ts
```diff
@@ -0,0 +1,302 @@
...
+  projectName: string | null;
+
+  @ApiProperty({
+    nullable: true,
+    description: "The associated project uuid"
+  })
+  projectUuid: string | null;
+
+  @ApiProperty({ nullable: true })
+  taskId: number | null;
```
roguenet, line 193: Is the numerical task ID useful in the FE? I would expect it to only care about the UUID.

apps/entity-service/src/entities/dto/site-report.dto.ts
```diff
@@ -102,8 +102,8 @@ export class SiteReportFullDto extends SiteReportLightDto {
       ...pickApiProperties(siteReport, SiteReportFullDto),
       lightResource: false,
       // these two are untyped and marked optional in the base model.
-      createdAt: siteReport.createdAt as Date,
-      updatedAt: siteReport.createdAt as Date,
+      createdAt: new Date(siteReport.createdAt),
+      updatedAt: new Date(siteReport.updatedAt),
```
roguenet, line 106: These really don't need the `new Date()` and I mentioned in my comment; please go back to `as Date`

apps/entity-service/src/entities/dto/site-report.dto.ts
roguenet: Everything else looks resolved, but this still needs attention.

roguenet (dismissed):

roguenet (approved):

PR #104, opened 2025-03-25
Fix/[TM-1852] button not filtering

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1852)
[Task Link](https://gfw.atlassian.net/browse/TM-1870)


roguenet (commented):
Should this really have so many code changes? I wouldn't expect something this big for a hotfix. Perhaps there's a 1 or 2 commits on the end that need to be rebased onto a branch from the hotfix instead of staging?

PR #105, opened 2025-03-25
[TM-1852] add missing values to filter and add q property to select s…

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1852)
[Task Link](https://gfw.atlassian.net/browse/TM-1870)

### DIFF
--- a/apps/entity-service/src/entities/dto/entity-query.dto.ts
+++ b/apps/entity-service/src/entities/dto/entity-query.dto.ts
@@ -26,7 +26,10 @@ export class EntityQueryDto extends IntersectionType(QuerySort, NumberPage) {
   @IsOptional()
   search?: string;
 
-  @ApiProperty({ required: false })
+  @ApiProperty({
+    required: false,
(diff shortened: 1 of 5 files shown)


apps/entity-service/src/entities/dto/entity-query.dto.ts
```diff
@@ -26,6 +26,10 @@ export class EntityQueryDto extends IntersectionType(QuerySort, NumberPage) {
   @IsOptional()
   search?: string;
 
+  @ApiProperty({ required: false })
+  @IsOptional()
+  q?: string;
```
roguenet, line 31: What is this? Could definitely use a better name, and if that's not possible, for sure needs a description.

roguenet: Hmm, I'm still not understanding why we need a second search query. But if it really is necessary, can we at least give it a better name? `filter` or `searchFilter` come to mind.

pachonjcl (approved):

roguenet (approved):

PR #107, opened 2025-03-25
[TM-1852] project organisation filter not working

LimberHope:
This initialization is being temporarily added to address the error 
"Property 'pageTotalFindOptions' does not exist on type 'PaginatedQueryBuilder<T>'."

### DIFF
--- a/libs/database/src/lib/util/paginated-query.builder.ts
+++ b/libs/database/src/lib/util/paginated-query.builder.ts
@@ -37,6 +37,7 @@ export class PaginatedQueryBuilder<T extends Model<T>> {
 
   constructor(private readonly MODEL: ModelCtor<T>, private readonly pageSize: number, include?: Includeable[]) {
     this.findOptions.limit = this.pageSize;
+    this.pageTotalFindOptions = {};
     if (include != null && include.length > 0) {
       this.findOptions.include = include;
     }


libs/database/src/lib/util/paginated-query.builder.ts
```diff
@@ -37,6 +37,7 @@ export class PaginatedQueryBuilder<T extends Model<T>> {
 
   constructor(private readonly MODEL: ModelCtor<T>, private readonly pageSize: number, include?: Includeable[]) {
     this.findOptions.limit = this.pageSize;
+    this.pageTotalFindOptions = {};
```
Scriptmatico, line 40: I don't feel this is the right solution, You are initializing this lines above:

`protected pageTotalFindOptions: FindOptions<Attributes<T>> = {};`

maybe is another kind of issue.

roguenet: Yeah, agreed; there must be something else going on because this property is already initialized when the class is created.

LimberHope: You're right, the pageTotalFindOptions property is already initialized above. However, I still haven't pinpointed the cause of the issue, as the error only occurs when deploying to the hotfix branch and not in my local environment. 🤔

PR #108, opened 2025-03-26
[TM-1773] Remove FrameworkUuid

Scriptmatico:
[Link to issue](https://gfw.atlassian.net/browse/TM-1773)


apps/entity-service/src/entities/processors/project.processor.ts
```diff
@@ -37,7 +37,6 @@ export class ProjectProcessor extends EntityProcessor<Project, ProjectLightDto,
     return await Project.findOne({
       where: { uuid },
       include: [
-        { association: "framework" },
```
roguenet, line 40: From the changes to the DTOs, it looks like this change is needed in the site and project report processors as well.

Scriptmatico: There is no association for project report processor. Already changed in site

roguenet: Oh neat; so this wasn't even working for project reports in the first place 😬

roguenet (approved):

PR #114, opened 2025-03-31
[TM-1862] Implement entity delete in v3 for all entities

pachonjcl:
[V3 PR](https://github.com/wri/terramatch-microservices/pull/114)
[BE PR](https://github.com/wri/wri-terramatch-api/pull/783)
[FE PR](https://github.com/wri/wri-terramatch-website/pull/1101)

[Task Link](https://gfw.atlassian.net/browse/TM-1862)

### DIFF
--- a/apps/entity-service/src/entities/entities.controller.ts
+++ b/apps/entity-service/src/entities/entities.controller.ts
@@ -17,7 +17,7 @@ import { PolicyService } from "@terramatch-microservices/common";
 import { buildDeletedResponse, buildJsonApi, getDtoType } from "@terramatch-microservices/common/util";
 import { SiteFullDto, SiteLightDto } from "./dto/site.dto";
 import { EntityIndexParamsDto } from "./dto/entity-index-params.dto";
-import { EntityQueryDto } from "./dto/entity-query.dto";
+import { EntityQueryDto, EntitySideload } from "./dto/entity-query.dto";
 import { MediaDto } from "./dto/media.dto";
 import { ProjectReportFullDto, ProjectReportLightDto } from "./dto/project-report.dto";
 import { NurseryFullDto, NurseryLightDto } from "./dto/nursery.dto";
@@ -27,7 +27,7 @@ import { NurseryReportFullDto, NurseryReportLightDto } from "./dto/nursery-repor
 import { SiteReportLightDto, SiteReportFullDto } from "./dto/site-report.dto";
 
 @Controller("entities/v3")
-@ApiExtraModels(ANRDto, ProjectApplicationDto, MediaDto)
+@ApiExtraModels(ANRDto, ProjectApplicationDto, MediaDto, EntitySideload)
 export class EntitiesController {
   constructor(private readonly policyService: PolicyService, private readonly entitiesService: EntitiesService) {}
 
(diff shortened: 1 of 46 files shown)


libs/common/src/lib/policies/nursery.policy.ts
```diff
@@ -23,6 +24,7 @@ export class NurseryPolicy extends UserPermissionsPolicy {
         ];
         if (projectIds.length > 0) {
           this.builder.can("read", Nursery, { projectId: { $in: projectIds } });
+          this.builder.can("delete", Nursery, { projectId: { $in: projectIds } });
```
roguenet, line 27: According to the PHP nursery soft delete controller, a nursery may only be deleted by a non-admin if there have been no reports generated. I was hoping we could implement that restriction in the policy, but I'm not thinking of a clean way to do it, so I think instead the nursery processor needs to overload `delete` and perform the check there.

libs/common/src/lib/policies/site.policy.ts
```diff
@@ -23,6 +24,7 @@ export class SitePolicy extends UserPermissionsPolicy {
         ];
         if (projectIds.length > 0) {
           this.builder.can("read", Site, { projectId: { $in: projectIds } });
+          this.builder.can("delete", Site, { projectId: { $in: projectIds } });
```
roguenet, line 27: The same restriction applies for non-admins deleting sites, so let's do the override in the site processor as well.

pachonjcl: changed accordingly

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -140,4 +149,17 @@ export class NurseryProcessor extends EntityProcessor<Nursery, NurseryLightDto,
       )[0].seedlingsYoungTrees ?? 0
     );
   }
+
+  async delete(nursery: Nursery, userId: number) {
+    const user = await User.findOne({ where: { id: userId }, include: [Role] });
```
roguenet, line 154: Processors can now get the user id through their entity service reference:
```
const user = await User.findOne({ where: { id: this.entitiesService.userId }, include: [Role] });
```

However, I don't think you should be fetching the user record here. See below.

roguenet (approved):

apps/entity-service/src/entities/entities.controller.ts
```diff
@@ -104,7 +104,7 @@ export class EntitiesController {
 
     await this.policyService.authorize("delete", model);
 
-    await processor.delete(model);
+    await processor.delete(model, this.policyService.userId);
```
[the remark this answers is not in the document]
pachonjcl: removed

apps/entity-service/src/entities/processors/nursery.processor.ts
pachonjcl: removed

apps/entity-service/src/entities/processors/nursery.processor.ts
```diff
@@ -140,4 +149,17 @@ export class NurseryProcessor extends EntityProcessor<Nursery, NurseryLightDto,
...
+    const isAdmin = user?.roles?.some(r => r.name.startsWith("admin"));
+    const reports = await NurseryReport.count({ where: { nurseryId: nursery.id } });
+
+    if (!isAdmin && reports > 0) {
+      throw new NotAcceptableException("You can only delete nurseries that do not have reports");
+    }
```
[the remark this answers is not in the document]
pachonjcl: changed using permissions

PR #122, opened 2025-04-09
[TM-1772] add media to entity association controller

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1772)

### DIFF
--- a/apps/entity-service/src/entities/dto/media.dto.ts
+++ b/apps/entity-service/src/entities/dto/media.dto.ts
@@ -1,14 +1,16 @@
 import { ApiProperty } from "@nestjs/swagger";
-import { JsonApiAttributes, pickApiProperties } from "@terramatch-microservices/common/dto/json-api-attributes";
+import { pickApiProperties } from "@terramatch-microservices/common/dto/json-api-attributes";
 import { Media } from "@terramatch-microservices/database/entities";
+import { AssociationDto, AssociationDtoAdditionalProps } from "./association.dto";
 
-export class MediaDto extends JsonApiAttributes<MediaDto> {
-  constructor(media: Media, url: string, thumbUrl: string) {
+export class MediaDto extends AssociationDto<MediaDto> {
+  constructor(media: Media, url: string, thumbUrl: string, additional: AssociationDtoAdditionalProps) {
     super({
       ...pickApiProperties(media, MediaDto),
       url,
       thumbUrl,
-      createdAt: media.createdAt
+      createdAt: media.createdAt,
+      ...additional
     });
   }
 
--- a/apps/entity-service/src/entities/entities.service.ts
+++ b/apps/entity-service/src/entities/entities.service.ts
@@ -63,6 +63,9 @@ const ASSOCIATION_PROCESSORS = {
       attributes: ["uuid", "name", "taxonId", "collection", [fn("SUM", col("amount")), "amount"]],
       group: ["taxonId", "name", "collection"]
     })
+  ),
+  medias: AssociationProcessor.buildSimpleProcessor(MediaDto, ({ id: modelId }, modelType) =>
+    Media.findAll({ where: { modelType, modelId } })
   )
 };
 
@@ -134,7 +137,11 @@ export class EntitiesService {
   fullUrl = (media: Media) => this.mediaService.getUrl(media);
   thumbnailUrl = (media: Media) => this.mediaService.getUrl(media, "thumbnail");
 
-  mediaDto = (media: Media) => new MediaDto(media, this.fullUrl(media), this.thumbnailUrl(media));
+  mediaDto = (media: Media) =>
+    new MediaDto(media, this.fullUrl(media), this.thumbnailUrl(media), {
+      entityType: media.modelType as EntityType,
+      entityUuid: media.modelId.toString()
+    });
 
   mapMediaCollection(media: Media[], collection: MediaCollection) {
     const grouped = groupBy(media, "collectionName");
--- a/apps/entity-service/src/entities/entity-associations.controller.ts
+++ b/apps/entity-service/src/entities/entity-associations.controller.ts
@@ -2,16 +2,18 @@ import { BadRequestException, Controller, Get, NotFoundException, Param } from "
 import { EntityAssociationIndexParamsDto } from "./dto/entity-association-index-params.dto";
 import { EntitiesService } from "./entities.service";
 import { PolicyService } from "@terramatch-microservices/common";
-import { buildJsonApi } from "@terramatch-microservices/common/util";
+import { buildJsonApi, DocumentBuilder } from "@terramatch-microservices/common/util";
 import { ApiExtraModels, ApiOperation } from "@nestjs/swagger";
 import { DemographicCollections, DemographicDto, DemographicEntryDto } from "./dto/demographic.dto";
 import { ExceptionResponse, JsonApiResponse } from "@terramatch-microservices/common/decorators";
 import { SeedingDto } from "./dto/seeding.dto";
 import { TreeSpeciesDto } from "./dto/tree-species.dto";
-
+import { MediaDto } from "./dto/media.dto";
+import { TMLogger } from "@terramatch-microservices/common/util/tm-logger";
 @Controller("entities/v3/:entity/:uuid")
 @ApiExtraModels(DemographicEntryDto, DemographicCollections)
 export class EntityAssociationsController {
+  private readonly logger = new TMLogger(EntityAssociationsController.name);
   constructor(private readonly entitiesService: EntitiesService, private readonly policyService: PolicyService) {}
 
   @Get(":association")
@@ -23,6 +25,8 @@ export class EntityAssociationsController {
     { data: DemographicDto, hasMany: true },
     { data: SeedingDto, hasMany: true },
     { data: TreeSpeciesDto, hasMany: true }
+    // TODO: check why this is causing a TypeError
+    // { data: MediaDto, hasMany: true }
   ])
   @ExceptionResponse(BadRequestException, { description: "Param types invalid" })
   @ExceptionResponse(NotFoundException, { description: "Base entity not found" })
@@ -30,8 +34,14 @@ export class EntityAssociationsController {
     const processor = this.entitiesService.createAssociationProcessor(entity, uuid, association);
     const baseEntity = await processor.getBaseEntity();
     await this.policyService.authorize("read", baseEntity);
-
-    const document = buildJsonApi(processor.DTO, { forceDataArray: true });
+    let document: DocumentBuilder;
+    this.logger.log(processor.DTO);
+    if (processor.DTO === MediaDto) {
+      // TODO: check why this is causing a TypeError
+      document = buildJsonApi(processor.DTO, { forceDataArray: true });
+    } else {
+      document = buildJsonApi(processor.DTO, { forceDataArray: true });
+    }
     await processor.addDtos(document);
     return document.serialize();
   }


apps/entity-service/src/entities/entities.service.ts
```diff
@@ -53,6 +55,10 @@ const ASSOCIATION_PROCESSORS = {
       Demographic.findAll({
         where: { demographicalType, demographicalId, hidden: false },
         include: [{ association: "entries" }]
+      }),
+    ({ id: demographicalId }, demographicalType) =>
+      Demographic.count({
+        where: { demographicalType: demographicalType.toString(), demographicalId, hidden: false }
```
roguenet, line 61: Unless I'm missing something, we only need this count accessor for models that intend to support pagination. It's possible we'll want to support pagination for demographics, but if we do, we'd want to save that for a future ticket.

apps/entity-service/src/entities/entities.service.ts
```diff
@@ -152,7 +159,14 @@ export class EntitiesService {
   fullUrl = (media: Media) => this.mediaService.getUrl(media);
   thumbnailUrl = (media: Media) => this.mediaService.getUrl(media, "thumbnail");
 
-  mediaDto = (media: Media) => new MediaDto(media, this.fullUrl(media), this.thumbnailUrl(media));
+  async mediaDto(media: Media) {
+    const userDto = await User.findOne({ where: { id: media.createdBy } });
```
roguenet, line 163: I would rename this to `user` to accurately reflect what the variable holds. Additionally, please use `attributes` to limit what gets pulled from the DB to just what the DTO needs.

A final tip: if this is going to be included in all direct media fetches, it would be better to include this association in the `media` fetch using the `include` array (and again, limiting the attributes to just what's needed for the user DTO). The way this is written now, it's an N+1 query because you're fetching all the Media, then fetching the user for each one. This is one of the main problems that is making the PHP API slow - the way that Laravel Eloquent is set up, it's _way too easy_ to descend into an N+1 query hole, and I want us to be especially cognizant of opportunities to intelligently reduce the number of SQL queries needed to satisfy an API request. We'll want to be especially aware of this pattern when speccing out the transition of Dashboard API requests to v3, as I believe this is one of the big bottlenecks in those controllers today.

If you'd like help wiring this up, let me know.

apps/entity-service/src/entities/processors/association-processor.ts
```diff
@@ -21,19 +26,26 @@ export abstract class AssociationProcessor<M extends UuidModel<M>, D extends Ass
    */
   static buildSimpleProcessor<M extends UuidModel<M>, D extends AssociationDto<D>>(
     dtoClass: Type<D>,
-    associationGetter: (entity: EntityModel, entityLaravelType: string) => Promise<M[]>
+    associationGetter: (entity: EntityModel, entityLaravelType: string, query?: MediaQueryDto) => Promise<M[]>,
+    totalGetter?: (entity: EntityModel, entityLaravelType: string, query?: MediaQueryDto) => Promise<number>
```
roguenet, line 30: Since media is the only one supporting pagination for now and it's now got its own subclass of `AssociationProcessor`, it might be a good idea to roll back this complication from the "simple" processor. I think it's reasonable for an association that needs pagination support to have its own processor subclass. The simpler processor was really just a way to make it easy to provide support for very simple associations.

apps/entity-service/src/entities/processors/association-processor.ts
```diff
@@ -53,20 +65,39 @@ export abstract class AssociationProcessor<M extends UuidModel<M>, D extends Ass
...
-    const associations = await this.getAssociations(await this.getBaseEntity());
+  async addDtos(document: DocumentBuilder, query?: MediaQueryDto): Promise<void> {
+    const associations = await this.getAssociations(await this.getBaseEntity(), query);
 
     const additionalProps = { entityType: this.entityType, entityUuid: this.entityUuid };
     const indexIds: string[] = [];
     for (const association of associations) {
       indexIds.push(association.uuid);
-      document.addData(association.uuid, new this.DTO(association, additionalProps));
+      if (this.DTO.name === MediaDto.name) {
```
roguenet, line 75: I definitely don't want to see custom code specific to one type of association DTO in the processor super class. Anything custom like this should be handled in the custom subclass.

libs/database/src/lib/entities/media.entity.ts
```diff
@@ -114,6 +114,22 @@ export class Media extends Model<Media> {
...
+      : this.modelType === Nursery.LARAVEL_TYPE
+      ? "nursery"
+      : this.modelType === Site.LARAVEL_TYPE
+      ? "site"
+      : this.modelType === NurseryReport.LARAVEL_TYPE
+      ? "nursery-report"
+      : this.modelType === ProjectReport.LARAVEL_TYPE
+      ? "project-report"
+      : this.modelType === SiteReport.LARAVEL_TYPE
+      ? "site-report"
```
roguenet, line 129: I'm not sure what these are being used for yet (maybe something in the FE?), but in general, v3 is trying to be consistent about how model names are specified. See `entities.ts` - v3 prefers plural, camelCased names.

libs/common/src/lib/decorators/transform-boolean-string.decorator.ts
```diff
@@ -0,0 +1,23 @@
+import { IsBoolean, IsOptional } from "class-validator";
+import { Transform } from "class-transformer";
+import { applyDecorators } from "@nestjs/common";
+
+type BooleanStreamOptions = {
```
roguenet, line 5: This is my mistake, but please rename this to `BooleanStringOptions`.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,230 @@
...
+    protected readonly entityModelClass: EntityClass<EntityModel>,
+    protected readonly entitiesService: EntitiesService
+  ) {
+    super(entityType, entityUuid, entityModelClass, entitiesService);
+  }
+
+  private async getSiteModels(site: Site) {
+    const models = [{ modelType: this.entityModelClass.LARAVEL_TYPE, ids: [site.id] }];
+
+    const siteReports = await SiteReport.findAll({ where: { siteId: site.id }, attributes: ["id"] });
```
roguenet, line 35: Why is fetching a site's media also fetching the media for the site report? That sort of pattern seems to be greatly complicating this processor. It's also not super intuitive to me. It seems like if I want the media for a site report, I'll send another request to get that site report's media. If it's super common to need all the media for a given site's reports on the site's page, then we should explore supporting that with a sideloading mechanism, like I did for the my projects page: https://github.com/wri/terramatch-microservices/blob/22c0727cdd850d169d02e9bbe14d1a26dba00017/apps/entity-service/src/entities/processors/project.processor.ts#L101-L112. If it seems from the FE usage like this is something we really need, I can help you add sideloading to the association processor pattern.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,230 @@
...
+    models.push({ modelType: SiteReport.LARAVEL_TYPE, ids: siteReports.map(report => report.id) });
+
+    return models;
```
roguenet, line 38: Lots of duplication in this class, and a little inefficiency here. In general, if you find yourself pulling a bunch of records just to get the ids to use in another query, that's a good candidate for using a subquery.

I see below that in most cases, this return result is being used to build an SQL `OR` where each element is specifying this laravel type and the set of IDs. I think instead, this should be using a general polymorphic subquery pattern. I've actually made that change to the Media entity on my entity update branch. I'm hoping to get that into PR tomorrow, but in the meantime, it would be easy enough to bring that change to this branch: https://github.com/wri/terramatch-microservices/blob/03865417d32b59fc741b62e3a4a46bddb8866374/libs/database/src/lib/entities/media.entity.ts#L43-L45.

Then (assuming we can do away with this loading of extra report media at this codepoint), instead of having all these extra methods for each type of media association, `buildQuery` below can do:
```
const builder = await this.entitiesService.buildQuery(Media.for(baseEntity), query);
```

apps/entity-service/src/entities/dto/media.dto.ts
```diff
@@ -1,14 +1,26 @@
...
+    thumbUrl: string,
+    userDto: UserDto,
+    additional: AssociationDtoAdditionalProps
+  ) {
     super({
       ...pickApiProperties(media, MediaDto),
       url,
       thumbUrl,
-      createdAt: media.createdAt
+      createdBy: userDto,
```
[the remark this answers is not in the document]
pachonjcl: moved to additional properties

apps/entity-service/src/entities/entities.service.ts
pachonjcl: removed.

apps/entity-service/src/entities/entities.service.ts
pachonjcl: moved as part of the media attributes

apps/entity-service/src/entities/processors/association-processor.ts
pachonjcl: removed

apps/entity-service/src/entities/processors/association-processor.ts
pachonjcl: removed as it wasn't needed after all

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: This one is tricky, because I need to be able to search and sort based on date of all records (for example by createdAt).
Watching the current side load I don't think this one fits.
Also this is the current behavior on v2

libs/common/src/lib/decorators/transform-boolean-string.decorator.ts
pachonjcl: changed

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: I think your PR is going to be merged today, will update this one accordingly.

apps/entity-service/src/entities/dto/association.dto.ts
```diff
@@ -1,9 +1,9 @@
 import { JsonApiAttributes } from "@terramatch-microservices/common/dto/json-api-attributes";
 
-export type AssociationDtoAdditionalProps = {
+export interface AssociationDtoAdditionalProps {
```
roguenet, line 3: Doesn't matter much, but I'm curious why you moved this to an interface?

apps/entity-service/src/entities/dto/media-association.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { Media } from "@terramatch-microservices/database/entities";
+import { AssociationDto, AssociationDtoAdditionalProps } from "./association.dto";
+import { UserDto } from "@terramatch-microservices/common/dto/user.dto";
+import { pickApiProperties } from "@terramatch-microservices/common/dto/json-api-attributes";
+export interface MediaAssociationDtoAdditionalProps extends AssociationDtoAdditionalProps {
```
roguenet, line 5: Please keep empty newlines between logical blocks to make it easier to visually parse a TS file. In general, I keep an empty line above all class, type (and interface), and method declarations.

Additionally, if the `AssociationDtoAdditionalProps` was moved to a interface just to make this `extends` possible, then please revert that and use the intersection operator instead:
```
export type MediaAssociationDtoAdditionalProps = AssociationDtoAdditionalProps & { 
```

apps/entity-service/src/entities/dto/association.dto.ts
pachonjcl: I wanted to extend this one for Media

apps/entity-service/src/entities/dto/media-association.dto.ts
pachonjcl: got it.

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: I don't quite get how to accomplish this one. Please can you suggest the change for this one?

apps/entity-service/src/entities/processors/media.processor.ts
roguenet: There are lots of bad behaviors on v2 that we don't want to replicate in v3. Which FE page is this needed for? Maybe I can take a look and suggest some alternative to product. It's totally possible the current behavior is not necessary and while we're moving things to v3 it's always important to identify and reduce bad code / API patterns.

apps/entity-service/src/entities/processors/media.processor.ts
roguenet: That suggestion above is basically it. The `for` method should now be available to you on `Media` and when creating the builder, instead of calling `buildQuery(Media, query)`, call `buildQuery(Media.for(baseEntity), query)`. That will limit the queries executed on by the builder to media that is associated with the given base entity.

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: `project/9db2495f-bee1-4323-97ac-c86a3a4e0283?tab=gallery`
`site/553f4d72-73a0-47b5-bc30-1389d243c5ce?tab=gallery`
Basically all the gallery tabs for PD views

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: I think this suggestion is related to previous comment, I will pause this one until we have some product feedback on the expected functionality for gallery, right?

roguenet: I'll do some investigation and follow up with product on this tomorrow. I have a meeting heavy day, but I'll make this  a priority for as soon as I'm able to devote time to it.

apps/entity-service/src/entities/entities.service.ts
```diff
@@ -157,7 +159,14 @@ export class EntitiesService {
   fullUrl = (media: Media) => this.mediaService.getUrl(media);
   thumbnailUrl = (media: Media) => this.mediaService.getUrl(media, "thumbnail");
 
-  mediaDto = (media: Media) => new MediaDto(media, this.fullUrl(media), this.thumbnailUrl(media));
+  mediaDto(media: Media) {
+    return new MediaDto(media, {
+      url: this.fullUrl(media),
+      thumbUrl: this.thumbnailUrl(media),
+      entityType: media.modelType as EntityType,
+      entityUuid: media.modelId.toString()
```
roguenet, line 167: A couple things I'm confused about here: 
* The DTO doesn't show there being these `entity` fields, nor does it seem to have entity UUID at all
* `modelType` is _not_ an `EntityType`, it's a Laravel model string (see my comment above about wanting to map those to an entity type
* `modelId` is not a UUID, it's a numerical int ID

apps/entity-service/src/entities/processors/media.processor.ts
roguenet: I did some digging into the FE implementation of these galleries. I don't love it, and I would probably implement the whole system differently if we were building from scratch, but I agree that these complicated multi-model queries are the way to go for this ticket.

I think there's room for improvement in a lot of this processor. I'm going to comment on a few items now.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+    models.push({ modelType: ProjectReport.LARAVEL_TYPE, subquery: projectReportSubquery.literal });
+
+    const siteSubquery = Subquery.select(Site, "id").eq("projectId", project.id);
+    models.push({ modelType: Site.LARAVEL_TYPE, subquery: siteSubquery.literal });
+
+    const sites = await Site.findAll({ where: { projectId: project.id }, attributes: ["id"] });
+
+    const siteReportSubquery = Subquery.select(SiteReport, "id").in(
+      "siteId",
+      sites.map(site => site.id)
```
roguenet, line 65: This should simply be `siteSubquery.literal`. The `Site.findAll()` should not be needed.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+    );
+    models.push({ modelType: SiteReport.LARAVEL_TYPE, subquery: siteReportSubquery.literal });
+
+    const nurseries = await Nursery.findAll({ where: { projectId: project.id }, attributes: ["id"] });
+    const nurserySubquery = Subquery.select(Nursery, "id").eq("projectId", project.id);
+    models.push({ modelType: Nursery.LARAVEL_TYPE, subquery: nurserySubquery.literal });
+
+    const nurseryReportSubquery = Subquery.select(NurseryReport, "id").in(
+      "nurseryId",
+      nurseries.map(nursery => nursery.id)
```
roguenet, line 75: Same here as above - fetching all nurseries should not be needed.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+    return models;
+  }
+
+  private async getProjectReportModels(projectReport: ProjectReport) {
+    const models: QueryModelType[] = [{ modelType: this.entityModelClass.LARAVEL_TYPE, subquery: [projectReport.id] }];
+
+    const projectReportWithDue = await ProjectReport.findOne({
+      where: { id: projectReport.id },
+      attributes: ["dueAt"]
+    });
```
roguenet, line 88: This is a very concerning pattern. We should definitely avoid re-fetching an already loaded model without a very good reason. If additional fields are needed, a good approach might be to have this processor override `getBaseEntity`. Or perhaps a better approach would be to create a method in `AssociationProcessor` that returns the list of attributes to retrieve. The base implementation would be:
```
/**
 * Returns all attributes that should be loaded on the base model load. 
 * 
 * Note: The code that uses this attribute will perform an intersection between this list and the available
 * attributes on the model class so it's OK to include attributes here that are not available on all Entity
 * classes.
 */
get baseModelAttributes() {
    // Only pull the attributes that are needed by the entity policies.
    return ["id", "frameworkKey", "projectId", "siteId", "nurseryId"];
}
```
In this processor, the implementation is:
```
get baseModelAttributes() {
  return [...super.baseModelAttributes, "dueAt"];
}
```
Then the usage in `getBaseEntity` is 
```
    const attributes = intersection(
      this.baseModelAttributes,
      Object.keys(this.entityModelClass.getAttributes())
    );
```

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const project = await Project.findOne({ where: { id: projectReport.projectId }, attributes: ["id"] });
```
roguenet, line 90: This is another red flag: I can't think of any reason why you'd need to go to the DB to only pull the ID field when you in fact already have the ID in memory. The two lines below can simply use `eq("projectId", projectReport.projectId)`

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    const siteSubquery = Subquery.select(Site, "id").eq("projectId", project.id);
+    const nurserySubquery = Subquery.select(Nursery, "id").eq("projectId", project.id);
+
+    let siteReports = [];
+    if (projectReportWithDue.dueAt) {
```
roguenet, line 96: Please use explicit null checks (`if (projectReport.dueAt != null) {`).

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+      siteReports = await SiteReport.findAll({
+        where: {
+          [Op.and]: [
+            { siteId: { [Op.in]: siteSubquery.literal } },
+            Sequelize.where(fn("MONTH", col("due_at")), projectReportWithDue.dueAt.getMonth() + 1),
+            Sequelize.where(fn("YEAR", col("due_at")), projectReportWithDue.dueAt.getFullYear())
```
roguenet, line 102: :art: Would be nice to get these capability into the SubqueryBuilder in order to avoid these loads, but it's going to be a little tricky and we can avoid it for now if you don't want to explore that.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+    } else if (baseEntity instanceof Site) {
+      models = await this.getSiteModels(baseEntity);
+    } else if (baseEntity instanceof Nursery) {
+      models = await this.getNurseryModels(baseEntity);
+    } else if (baseEntity instanceof ProjectReport) {
+      models = await this.getProjectReportModels(baseEntity);
+    } else if (baseEntity instanceof SiteReport) {
+      models = await this.getSiteReportModels(baseEntity);
+    } else if (baseEntity instanceof NurseryReport) {
+      models = [{ modelType: this.entityModelClass.LARAVEL_TYPE, subquery: [baseEntity.id] }];
```
roguenet, line 153: the `getSiteReportModels` method is doing this exact same thing. I'd either remove that method entirely, or rename it to something like `getBaseEntityModels()` and use it in both cases.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+              [query.isGeotagged ? Op.ne : Op.eq]: null
+            }
+          }
+        ]
+      });
+    }
+
+    if (query.isPublic != null) {
+      builder.where({
+        isPublic: query.isPublic ? "1" : "0"
```
roguenet, line 184: `isPublic` on the model is boolean, this should simply be `isPublic: query.isPublic`.

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -0,0 +1,245 @@
...
+        entityType: this.entityType,
+        entityUuid: this.entityUuid,
+        url: this.entitiesService.fullUrl(media),
+        thumbUrl: this.entitiesService.thumbnailUrl(media)
+      };
+
+      document.addData(association.uuid, new this.DTO(association, additionalProps));
+    }
+
+    const total = await this.getTotal(await this.getBaseEntity(), query);
```
roguenet, line 234: Between this line and the `getAssociations` call above, you're calling `buildQuery()` twice, which is not a cheap operation. This solution I think is to: 
* add memoization to `buildQuery()`, similar to what's in place on `getBaseEntity`. 
* rename to `private async getQueryBuilder()`
  * Note that this means `getQueryBuilder` needs access to the base entity and query. It can get the base entity by calling `this.getBaseEntity()` itself, as for the query, I think that passing it to this method is already a little weird. Let's instead pass it to `createAssociationProcessor()` and then on to the processor constructor so it becomes a local property of the processor object.

I would then also remove the `getTotal()` method, and this line simply becomes:
```
const total = await (await this.getQueryBuilder()).paginationTotal();
```

libs/database/src/lib/entities/media.entity.ts
```diff
@@ -116,6 +117,22 @@ export class Media extends Model<Media> {
   @Column(BIGINT.UNSIGNED)
   createdBy: number | null;
 
+  @BelongsTo(() => User)
```
roguenet, line 120: Please add `constraints: false`:
```
@BelongsTo(() => User, { constraints: false })
```

pachonjcl: added

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: changed

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: fixed

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: implemented.

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: you are right, removed

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: changed, please take a look.

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: let's avoid this one for now

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: removed too

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: changed

apps/entity-service/src/entities/dto/media.dto.ts
```diff
@@ -56,4 +59,13 @@ export class MediaDto extends JsonApiAttributes<MediaDto> {
 
   @ApiProperty({ nullable: true })
   photographer: string | null;
+
+  @ApiProperty({ nullable: true })
+  modelType: string | null;
+
+  @ApiProperty({ nullable: true })
+  createdBy: number | null;
```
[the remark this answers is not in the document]
pachonjcl: removed

apps/entity-service/src/entities/entities.service.ts
pachonjcl: changed

apps/entity-service/src/entities/processors/media.processor.ts
```diff
@@ -225,21 +217,31 @@ export class MediaProcessor extends AssociationProcessor<Media, MediaDto> {
...
   }
+
+  _queryBuilder: Promise<PaginatedQueryBuilder<Media>> | null = null;
+
+  get queryBuilder() {
+    if (this._queryBuilder == null) {
+      this._queryBuilder = this.getQueryBuilder();
+    }
+    return this._queryBuilder;
+  }
```
roguenet, line 246: I don't love a getter that returns a promise - getters typically should return a resolved value. This is also not following previous patterns in this codebase. 

Instead, `getQueryBuilder()` should do the memoization itself and simply return the builder if it has already been built. See how this pattern is established in `getBaseEntity`

apps/entity-service/src/entities/dto/media-association.dto.ts
```diff
@@ -1,17 +1,23 @@
...
 };
 
 export class MediaAssociationDto extends AssociationDto<MediaAssociationDto> {
   constructor(media: Media, additional: MediaAssociationDtoAdditionalProps) {
     super({
       ...pickApiProperties(media, MediaAssociationDto),
-      ...additional
+      ...additional,
+      entityType: additional.entityType,
+      entityUuid: additional.entityUuid
```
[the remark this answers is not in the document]
pachonjcl: unfortunately it is because of typescript, I just updated.

apps/entity-service/src/entities/processors/media.processor.ts
pachonjcl: changed

apps/entity-service/src/entities/dto/media.dto.ts
```diff
@@ -56,4 +59,13 @@ export class MediaDto extends JsonApiAttributes<MediaDto> {
 
   @ApiProperty({ nullable: true })
   photographer: string | null;
+
+  @ApiProperty({ nullable: true })
+  modelType: string | null;
```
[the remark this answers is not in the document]
pachonjcl: ok, this one is gone now

apps/entity-service/src/entities/dto/media.dto.ts
```diff
@@ -11,6 +11,10 @@ export class MediaDto extends AssociationDto<MediaDto> {
     super({
       ...pickApiProperties(media, MediaDto),
       ...additional,
+      entityType: additional.entityType,
+      entityUuid: additional.entityUuid,
+      url: additional.url,
+      thumbUrl: additional.thumbUrl,
```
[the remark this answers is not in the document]
roguenet: I think this is not the right solution. If you look at the particulars of the error that TS gives you, it's legitimately flagging that `url` is optional in the additional props, but required in the DTO. That's a real error, and it should be addressed appropriately with an update to the DTO, or a default value.

image

apps/entity-service/src/entities/dto/media-association.dto.ts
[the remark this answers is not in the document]
pachonjcl: changed

apps/entity-service/src/entities/dto/media.dto.ts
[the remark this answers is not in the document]
pachonjcl: changed

roguenet (approved):

PR #129, opened 2025-04-15
[TM-1915] Switch to using the project landscape string instead of the expensive geojson intersect query.

roguenet:
https://gfw.atlassian.net/browse/TM-1915


roguenet:
Moved this to draft; after consulting with Rhiannon, I need to update it so that `projectCohort` and `landscape` can be used together in the query.

pachonjcl (approved):

PR #137, opened 2025-04-18
[TM-1937] Add nothing report column to reports

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1937

### DIFF
--- a/apps/entity-service/src/entities/dto/nursery-report.dto.ts
+++ b/apps/entity-service/src/entities/dto/nursery-report.dto.ts
@@ -89,6 +89,9 @@ export class NurseryReportLightDto extends EntityDto {
 
   @ApiProperty()
   createdAt: Date;
+
+  @ApiProperty({ nullable: true })
+  nothingToReport: boolean | null;
 }
 
 export type AdditionalNurseryReportLightProps = Pick<NurseryReportLightDto, "reportTitle">;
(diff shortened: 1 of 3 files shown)


libs/database/src/lib/constants/status.ts
```diff
@@ -62,7 +62,7 @@ export const ReportStatusStates: States<ReportModel, ReportStatus> = {
   transitionValidForModel: (from: ReportStatus, to: ReportStatus, report: ReportModel) => {
     if ((from === DUE && to === AWAITING_APPROVAL) || (from === AWAITING_APPROVAL && to === STARTED)) {
       // these two transitions are only allowed for site / nursery reports when the nothingToReport flag is true;
-      return !(report instanceof ProjectReport) && report.nothingToReport;
+      return !(report instanceof ProjectReport) && (report.nothingToReport || false);
```
roguenet, line 65: I'm curious about this change. Two things:
* Why was it needed? 
* For me, this is hard to read at a glance. Something like this would be better IMO: `&& report.nothingToReport === true`, or indicating that it's a real default with ?? instead: `&& (report.nothingToReport ?? false)`

Scriptmatico: Original code is throwing this error due the optional parameter
image

But I agree with you probably === true is better

roguenet: Weird... I feel like the build should have failed when I merged this code in, but it didn't 🤷🏻

roguenet (approved):

PR #142, opened 2025-04-23
[TM-1922] sideload associations

egrojMonroy:


apps/entity-service/src/entities/processors/project-report.processor.ts
```diff
@@ -128,6 +132,30 @@ export class ProjectReportProcessor extends ReportProcessor<
...
+      const serialized = entityDocument.serialize();
+      if (serialized.data) {
+        if (Array.isArray(serialized.data)) {
+          serialized.data.forEach(resource => {
+            document.addIncluded(resource.id, resource.attributes);
+          });
+        } else {
+          document.addIncluded(serialized.data.id, serialized.data.attributes);
+        }
+      }
```
roguenet, line 153: What's the reason for the intermediate document builder / serialize call? That's certainly not how the builder was intended to be used... let's find a better solution, but I need to know why you think you needed it first.

egrojMonroy: I will search another solution.
But I added this serialized because AssociationProcessor.addDtos() can added data to main data elements, while sideloading needed the data as included.

egrojMonroy: I added another function. 
Maybe is this a better solution ?

apps/entity-service/src/entities/processors/association-processor.ts
```diff
@@ -70,4 +70,22 @@ export abstract class AssociationProcessor<M extends UuidModel, D extends Associ
...
   }
+
+  async addIncludedDtos(document: DocumentBuilder): Promise<void> {
+    const associations = await this.getAssociations(await this.getBaseEntity());
+
+    const additionalProps = { entityType: this.entityType, entityUuid: this.entityUuid };
+    const indexIds: string[] = [];
+    for (const association of associations) {
+      indexIds.push(association.uuid);
+      document.addIncluded(association.uuid, new this.DTO(association, additionalProps));
```
roguenet, line 81: The only difference between these two methods is this call - please unify to a single method with a flag in the args:
```
async addDtos(document: DocumentBuilder, asIncluded = false) {
```

egrojMonroy: indeed, it is unified now. I should have catched this overlap. appreciate the catch!

roguenet (approved):

PR #145, opened 2025-04-25
[TM-1959] Move get Disturbances Stratas, Invasives associations

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1959


apps/entity-service/src/entities/dto/disturbance.dto.ts
```diff
@@ -0,0 +1,39 @@
...
+  extent: string | null;
+
+  @ApiProperty({ nullable: true })
+  description: string | null;
+
+  @ApiProperty({ nullable: true })
+  oldId: number;
+
+  @ApiProperty({ nullable: true })
+  oldModel: string | null;
```
roguenet, line 35: Are `oldId` and `oldModel` used in the FE? I feel like we might be able to get rid of them.

apps/entity-service/src/entities/dto/invasive.dto.ts
```diff
@@ -0,0 +1,30 @@
...
+  type: string | null;
+
+  @ApiProperty({ nullable: true })
+  name: string | null;
+
+  @ApiProperty({ nullable: true })
+  oldId: number;
+
+  @ApiProperty({ nullable: true })
+  oldModel: string | null;
```
roguenet, line 26: Same question here re `oldId` / `oldModel`

apps/entity-service/src/entities/entities.service.ts
```diff
@@ -68,6 +80,17 @@ const ASSOCIATION_PROCESSORS = {
...
+  disturbances: AssociationProcessor.buildSimpleProcessor(
+    DisturbanceDto,
+    ({ id: disturbanceableId }, disturbanceableType) =>
+      Disturbance.findAll({ where: { disturbanceableType, disturbanceableId, hidden: false } })
+  ),
+  invasives: AssociationProcessor.buildSimpleProcessor(InvasiveDto, ({ id: invasiveableId }, invasiveableType) =>
+    Invasive.findAll({ where: { invasiveableType, invasiveableId, hidden: false } })
+  ),
+  stratas: AssociationProcessor.buildSimpleProcessor(StrataDto, ({ id: stratasableId }, stratasableType) =>
+    Strata.findAll({ where: { stratasableType, stratasableId, hidden: false } })
```
roguenet, line 93: Since these are all explicitly only returning non-hidden entries, it seems like we should be able to drop `hidden` from the DTOs as well.

apps/entity-service/src/entities/entities.service.ts
```diff
@@ -96,7 +119,10 @@ export class EntitiesService {
     await this.policyService.authorize(action, subject);
   }
 
-  async isFrameworkAdmin<T extends EntityModel>({ frameworkKey }: T) {
+  async isFrameworkAdmin<T extends EntityModel>(T: EntityModel) {
+    if (T instanceof Disturbance) return;
+    if (T instanceof Invasive) return;
+    const { frameworkKey } = T;
```
roguenet, line 125: A few problems here:
* You've changed the type of the model accepted by this function - it is meant to accept a value of type T, not base `EntityModel`. The param to the method should be something more like `(entityModel: T)`
* Disturbance and Invasive are not entity models, so I'm not really sure what's going on here.
* You've chagned the return type of the method from `boolean` to `boolean | undefined`. If this change is really necessary, the two early returns should explicitly return a boolean.

libs/database/src/lib/entities/stratas.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  @Column(BIGINT.UNSIGNED)
+  override id: number;
+
+  @Index
+  @Column({ type: UUID, defaultValue: UUIDV4 })
+  uuid: string;
+
+  @AllowNull
+  @Column(INTEGER.UNSIGNED)
+  owner_id: number | null;
```
roguenet, line 17: `ownerId`

libs/database/src/lib/entities/disturbance.entity.ts
```diff
@@ -40,14 +40,6 @@ export class Disturbance extends Model<Disturbance> {
   @Column(TEXT)
   description: string | null;
 
-  @AllowNull
-  @Column(INTEGER.UNSIGNED)
-  oldId: number;
-
-  @AllowNull
-  @Column(STRING)
-  oldModel: string | null;
```
roguenet: In cases like this, I think we want the Entity to still match what's in the DB. I've been adding a `@deprecated` flag with a comment to indicate that we can likely drop the column later. Lots of examples in [Media](https://github.com/wri/terramatch-microservices/blob/03865417d32b59fc741b62e3a4a46bddb8866374/libs/database/src/lib/entities/media.entity.ts#L120)

roguenet (approved):

PR #153, opened 2025-05-05
[TM-1981] remove plant end

egrojMonroy:


libs/database/src/lib/entities/site-polygon.entity.ts
```diff
@@ -106,10 +106,6 @@ export class SitePolygon extends Model<SitePolygon> {
   @Column({ type: DATE, field: "plantstart" })
   plantStart: Date | null;
 
-  @AllowNull
-  @Column({ type: DATE, field: "plantend" })
-  plantEnd: Date | null;
```
roguenet, line 111: It appears that the PHP PR isn't removing this column from the DB. We want these entities to reflect what's in the actual DB, even if it isn't currently being used. So one of two things should happen:
* Add a migration to the PHP PR to drop this column
* Or bring this back here and add a comment something along the lines of `/** @deprecated this column is no longer in use */`

roguenet (approved):

PR #156, opened 2025-05-06
[TM-1995] Migrate project pitches be v3

Scriptmatico:
https://gfw.atlassian.net/browse/TM-1995


apps/entity-service/src/entities/dto/project-pitch.dto.ts
```diff
@@ -0,0 +1,470 @@
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { ApiProperty } from "@nestjs/swagger";
+import { IsArray, IsDate, IsNumber, IsOptional, IsString, IsUUID, Max, Min } from "class-validator";
+import { JsonApiAttributes, pickApiProperties } from "@terramatch-microservices/common/dto/json-api-attributes";
+import { ProjectPitch } from "@terramatch-microservices/database/entities";
+
+@JsonApiDto({ type: "projectPitches" })
+export class ProjectPitchDto extends JsonApiAttributes<ProjectPitchDto> {
+  constructor(data: ProjectPitch) {
+    pickApiProperties(data, ProjectPitchDto);
```
roguenet, line 10: This call shouldn't be needed with the call to super below. This is assigning the properties to the DTO object twice.

apps/entity-service/src/entities/dto/project-pitch.dto.ts
```diff
@@ -0,0 +1,470 @@
...
+    super({
+      ...pickApiProperties(data, ProjectPitchDto)
+    });
+  }
+
+  @ApiProperty()
+  id: number;
```
roguenet, line 17: Is this ID needed on the client? We try to avoid sending numerical IDs when possible, and they usually aren't.

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    const projectPitch = await ProjectPitch.findOne({ where: { uuid } });
+    if (!projectPitch) {
+      throw new NotFoundException("ProjectPitch not found");
+    }
+    return projectPitch;
+  }
+
+  async getProjectPitches(userId: number, params: EntityQueryDto) {
+    const user = await User.findOne({
+      include: ["organisations"],
```
roguenet, line 20: Organisations is a bit table; please limit this to the attributes you need. Looks like it's just UUID

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -0,0 +1,79 @@
...
+      where: { id: userId }
+    });
+
+    if (!user) {
```
roguenet, line 24: Please use explicit null checks (`user == null`)

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    builder.where({
+      organisationId: {
+        [Op.in]: user.organisations.map(org => org.uuid)
+      }
+    });
+
+    return { data: await builder.execute(), paginationTotal: await builder.paginationTotal(), pageNumber };
+  }
+
+  async getAdminProjectPitches(params: EntityQueryDto) {
```
roguenet, line 57: There should be only one endpoint / function for fetching pitches. In the case of this service, the function above should simply check if the user has any `framework-` permissions, and return all pitches if it does. It's not perfect, but this system is going to see some changes when we move the form system over to v3, so I think it's good enough for now.

apps/entity-service/src/entities/project-pitches.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  @Get()
+  @ApiOperation({
+    operationId: "ProjectPitchesIndex",
+    summary: "Get projects pitches."
+  })
+  @JsonApiResponse([{ data: ProjectPitchDto, pagination: "number" }])
+  @ExceptionResponse(BadRequestException, { description: "Param types invalid" })
+  @ExceptionResponse(NotFoundException, { description: "Records not found" })
+  async getPitches(@Request() { authenticatedUserId }, @Query() params: EntityQueryDto) {
```
roguenet, line 31: These methods and operationIds should follow the conventions set by the entity service. In this case, this controller should have `projectPitchIndex` and `projectPitchGet`, with the operation ID matching the function name. 

Additionally, a new project pitch policy is needed that matches the the read permissions in the policy from PHP. Let me know know if you need help wiring that up.

Let's also have a query DTO specific to this endpoint. The entity one is already overloaded, and is not meant to be a generic grab bag of whatever query params are needed for all entity service endpoints.

Finally, I see in the admin controller in PHP that there are lots of filtering and sorting controls needed for the admin dashboard that do not appear to be implemented here.

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -36,44 +27,48 @@ export class ProjectPitchService {
...
-
-    if (params.search) {
-      builder.where({
-        [Op.or]: [
-          { projectName: { [Op.like]: `%${params.search}%` } },
-          { "$organisation.name$": { [Op.like]: `%${params.search}%` } }
-        ]
+    if (query.filter) {
+      Object.keys(query.filter).forEach(key => {
+        if (!["restoration_intervention_types", "project_country"].includes(key)) {
```
roguenet, line 40: These should be camel case.

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -36,44 +27,48 @@ export class ProjectPitchService {
...
+          "project_objectives",
+          "project_country",
+          "project_county_district",
+          "restoration_intervention_types",
+          "total_hectares",
+          "total_trees",
+          "capacity_building_needs",
+          "created_at",
+          "updated_at",
+          "deleted_at"
```
roguenet, line 64: These sort fields should all be camel case. Also, why is `deleted_at` here? There's no way for the FE to receive deleted entities typically, so sorting on that field makes no sense. Also `id`? That should be the default sort, and isn't typically needed as an explicit sort from the FE - it's a bit weird for the FE to request sorting on a field that it doesn't receive. In general, please make sure these are all actually needed in the places pitches are displayed. There's usually only something like 2-5 fields that we really need to sort an index on.

apps/entity-service/src/entities/project-pitches.controller.ts
```diff
@@ -1,66 +1,34 @@
...
-    const document = buildJsonApi(ProjectPitchDto, { pagination: "number" });
-    const indexIds: string[] = [];
-    for (const pitch of data) {
-      indexIds.push(pitch.uuid);
-      const pitchDto = new ProjectPitchDto(pitch);
-      document.addData(pitchDto.uuid, pitchDto);
+  async projectPitchIndex(@Query() params: ProjectPitchQueryDto) {
+    const permissions = await this.policyService.getPermissions();
+    if (!permissions.some(permission => permission.startsWith("framework-"))) {
+      throw new BadRequestException("User does not have permission to access this resource");
```
roguenet, line 29: This is incorrect. Pulling permissions allows the endpoint to filter which resources are fetched from the DB. However, there needs to be a policy in place that explicitly lists which resources the current user is allowed to access, and how they're allowed to access them. For this PR, you only need worry about `read`, but it still needs to be there. 

See my in-progress task branch for an example that's similar to what you need here:
* Policy file: https://github.com/wri/terramatch-microservices/blob/feat/TM-2004-v3-tasks/libs/common/src/lib/policies/task.policy.ts
* Checking the policy in the controller (also some examples of only pulling the correct entities based on what we expect the user to have access to): https://github.com/wri/terramatch-microservices/blob/feat/TM-2004-v3-tasks/apps/entity-service/src/entities/tasks.controller.ts#L95

apps/entity-service/src/entities/project-pitches.controller.ts
```diff
@@ -24,17 +24,16 @@ export class ProjectPitchesController {
...
-    }
     const { data, paginationTotal, pageNumber } = await this.projectPitchService.getProjectPitches(params);
     const document = buildJsonApi(ProjectPitchDto, { pagination: "number" });
     const indexIds: string[] = [];
-    for (const pitch of data) {
-      indexIds.push(pitch.uuid);
-      const pitchDto = new ProjectPitchDto(pitch);
-      document.addData(pitchDto.uuid, pitchDto);
+    if (data.length !== 0) {
+      await this.policyService.authorize("read", data);
```
roguenet, line 31: I'm glad the index is using the policy now, but the individual GET endpoint below needs it as well.

libs/common/src/lib/policies/project-pitch.policy.ts
```diff
@@ -0,0 +1,11 @@
+import { ProjectPitch } from "@terramatch-microservices/database/entities";
+import { UserPermissionsPolicy } from "./user-permissions.policy";
+
+export class ProjectPitchPolicy extends UserPermissionsPolicy {
+  async addRules() {
+    if (this.permissions.some(permission => permission.startsWith("framework-"))) {
+      this.builder.can("read", ProjectPitch);
+      return;
+    }
```
roguenet, line 9: This seems to be missing permissions for any PD levels. I haven't double-checked in the FE PR yet; have we removed any need for PDs to directly fetch pitches? If so, this is OK, but if not we need to address that.

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -0,0 +1,62 @@
...
+import { Includeable, Op } from "sequelize";
+import { PaginatedQueryBuilder } from "@terramatch-microservices/database/util/paginated-query.builder";
+import { MAX_PAGE_SIZE } from "./entities.service";
+import { ProjectPitchQueryDto } from "./dto/project-pitch-query.dto";
+
+@Injectable()
+export class ProjectPitchService {
+  async getProjectPitch(uuid: string) {
+    const projectPitch = await ProjectPitch.findOne({ where: { uuid } });
+    if (!projectPitch) {
```
roguenet, line 12: Please remember to use explicit null checks: `if (projectPitch == null)`

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -0,0 +1,62 @@
...
+      throw new NotFoundException("ProjectPitch not found");
+    }
+    return projectPitch;
+  }
+
+  async getProjectPitches(query: ProjectPitchQueryDto) {
+    const pageNumber = query.page ? query.page.number : 1;
+    const pageSize = query.page ? query.page.size : MAX_PAGE_SIZE;
```
roguenet, line 20: This construct doesn't handle the case of `query.page.size` or `query.page.number` being null, and doesn't do any validation on the resulting values. It's also using implicit null checks (`query.page ?` instead of `query.page != null ?`). I like how the entities service does it:
```
    const { size: pageSize = MAX_PAGE_SIZE, number: pageNumber = 1 } = page ?? {};
    if (pageSize > MAX_PAGE_SIZE || pageSize < 1) {
      throw new BadRequestException(`Invalid page size: ${pageSize}`);
    }
    if (pageNumber < 1) {
      throw new BadRequestException(`Invalid page number: ${pageNumber}`);
    }
```

apps/entity-service/src/entities/project-pitch.service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    if (query.search) {
+      builder.where({
+        [Op.or]: [
+          { projectName: { [Op.like]: `%${query.search}%` } },
+          { "$organisation.name$": { [Op.like]: `%${query.search}%` } }
+        ]
+      });
+    }
+    if (query.filter) {
```
roguenet, line 38: This one and the one line 30 should be using `!= null`

libs/common/src/lib/policies/project-pitch.policy.ts
```diff
@@ -0,0 +1,11 @@
+import { ProjectPitch } from "@terramatch-microservices/database/entities";
+import { UserPermissionsPolicy } from "./user-permissions.policy";
+
+export class ProjectPitchPolicy extends UserPermissionsPolicy {
+  async addRules() {
+    if (this.permissions.some(permission => permission.startsWith("framework-"))) {
```
roguenet, line 6: This can be simplified with a handy getter from `UserPermissionsPolicy`: 
```
if (this.frameworks.length > 0) {
```

PR #157, opened 2025-05-07
[TM-2008] add initial dashboard service

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-2008)
[Task Link](https://gfw.atlassian.net/browse/TM-2056)


apps/dashboard-service/src/app.module.ts
```diff
@@ -0,0 +1,35 @@
...
+
+@Module({
+  imports: [
+    SentryModule.forRoot(),
+    CommonModule,
+    HealthModule,
+    RedisModule.forRoot({
+      type: "single",
+      url: "redis://127.0.0.1:6379"
+    }),
```
roguenet, line 21: If you're going to need direct Redis access, please follow the pattern from the Data API: https://github.com/wri/terramatch-microservices/blob/2fb4a3b02b24bc697784d19af002f80825fc5d27/libs/data-api/src/lib/data-api.module.ts#L9-L19

apps/dashboard-service/src/dashboard/dto/cache.service.ts
```diff
@@ -0,0 +1,83 @@
+import { Injectable } from "@nestjs/common";
+import { Queue } from "bullmq";
+import { InjectQueue } from "@nestjs/bullmq";
+import { DashboardQueryDto } from "./dashboard-query.dto";
+
+import Redis from "ioredis";
+import Cache from "ioredis-cache";
```
roguenet, line 7: Ooh neat, I didn't find this library when I was doing the data api integration, and just rolled my own cache. I'll have to look into using it there 👍🏻

apps/dashboard-service/src/dashboard/dto/cache.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+@Injectable()
+export class CacheService {
+  private cache: Cache;
+
+  constructor(@InjectQueue("dashboard") private readonly dashboardQueue: Queue) {
+    this.cache = new Cache(new Redis());
```
roguenet, line 14: is `new Redis()` working? I think you'll actually want: 
```
@InjectRedis() private readonly redis: Redis
```

apps/dashboard-service/src/dashboard/dto/cache.service.ts
roguenet: Although now that I read closer; are you actually using this cache library? It looks like all your uses are just pulling the redis instance and using it directly.

apps/dashboard-service/src/dashboard/dto/delayed-job.dto.ts
```diff
@@ -0,0 +1,71 @@
+import { JsonApiAttributes } from "@terramatch-microservices/common/dto/json-api-attributes";
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { ApiProperty } from "@nestjs/swagger";
+
+const STATUSES = ["pending", "failed", "succeeded"];
+type Status = (typeof STATUSES)[number];
+
+@JsonApiDto({ type: "delayedJobs" })
+export class DelayedJobDto extends JsonApiAttributes<DelayedJobDto> {
```
roguenet, line 9: Why is the dashboard getting its own delayed job implementation?

apps/dashboard-service/src/dashboard/dto/total-section-header.service.ts
```diff
@@ -0,0 +1,93 @@
...
+  SiteReport,
+  TreeSpecies
+} from "@terramatch-microservices/database/entities";
+import { Op } from "sequelize";
+import { DashboardQueryDto } from "./dashboard-query.dto";
+import { DashboardProjectsQueryBuilder } from "@terramatch-microservices/database/util/dashboard-query.builder";
+
+@Injectable()
+export class TotalSectionHeaderService {
+  constructor() {}
```
roguenet, line 18: If the constructor will remain empty, best to just remove it.

apps/dashboard-service/src/dashboard/total-section-header.controller.ts
```diff
@@ -0,0 +1,27 @@
+import { Get, Query, Controller } from "@nestjs/common";
+import { ApiOperation } from "@nestjs/swagger";
+import { DashboardQueryDto } from "./dto/dashboard-query.dto";
+import { JsonApiResponse } from "@terramatch-microservices/common/decorators";
+import { DelayedJobDto } from "./dto/delayed-job.dto";
+import { CacheService } from "./dto/cache.service";
+import { DelayedJob } from "@terramatch-microservices/database/entities";
+
+@Controller("v3/dashboard/totalSectionHeader")
```
roguenet, line 9: Please use plurals for resource paths. Also, this needs a namespace (the bit before `v3`). This should probably be something more like `@Controller("dashboard/v3/totalSectionHeaders")`

apps/dashboard-service/src/dashboard/total-section-header.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+  async getTotalSectionHeader(@Query() query: DashboardQueryDto) {
+    const cacheKey = `dashboard:total-section-header|${this.cacheService.getCacheKeyFromQuery(query)}`;
+    const cachedData = await this.cacheService.get(cacheKey);
+    if (cachedData == null) {
+      const delayedJob = await DelayedJob.create({});
+      await this.cacheService.getTotalSectionHeader(cacheKey, query, delayedJob.id);
+      const delayedJobDto = new DelayedJobDto(delayedJob);
+      return delayedJobDto;
+    }
+    return JSON.parse(cachedData);
```
roguenet, line 25: Every v3 controller should be returning a JSON API structure (typically by using the `buildJsonApi()` builder). Also, rather than returning the delayed job DTO, it should return the delayed job ID, and let the FE delayed job connection / middleware (which we'll need to add an integration for in v3) handle the polling of the delayed job, similar to how it's done in v2.

apps/dashboard-service/src/dashboard/worker/worker.processor.ts
```diff
@@ -0,0 +1,30 @@
+import { Processor, WorkerHost } from "@nestjs/bullmq";
+import { TotalSectionHeaderService } from "../dto/total-section-header.service";
+import { Job } from "bullmq";
+import { TMLogger } from "@terramatch-microservices/common/util/tm-logger";
+import { DashboardQueryDto } from "../dto/dashboard-query.dto";
+import { CacheService } from "../dto/cache.service";
+import { DelayedJob } from "@terramatch-microservices/database/entities";
+@Processor("dashboard")
+export class WorkerProcessor extends WorkerHost {
```
roguenet, line 9: This could use a more specific name. `DashboardProcessor`, or better yet, making this system more generic and putting in in the `common` lib with something like `DelayedJobProcesor`. This won't be the only service that needs to process delayed jobs in the end.

libs/database/src/lib/entities/project-report.entity.ts
```diff
@@ -96,6 +96,15 @@ export class ProjectReport extends Model<ProjectReport> {
     return builder.literal;
   }
 
+  static approvedIdsSubqueryProjects(projectIds: number[], opts: ApprovedIdsSubqueryOptions = {}) {
+    const builder = Subquery.select(ProjectReport, "id")
+      .in("projectId", projectIds)
+      .in("status", ProjectReport.APPROVED_STATUSES);
+    if (opts.dueAfter != null) builder.gte("dueAt", opts.dueAfter);
+    if (opts.dueBefore != null) builder.lt("dueAt", opts.dueBefore);
```
roguenet, line 104: It looks like nothing is using this due after / before option currently. Unless you know you're going to need it, let's not add it for now.

roguenet (commented):
I didn't read in super deep detail, but called out a few things that stood out to me. No major red flags.

apps/dashboard-service/src/dashboard/dto/total-section-header.service.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  async getTotalHectaresSum(projectsIds: number[]) {
+    if (projectsIds.length === 0) {
+      return 0;
+    }
+    const totalHectaresRestoredSum =
+      (await SitePolygon.active().approved().sites(Site.approvedUuidsProjectsSubquery(projectsIds)).sum("calcArea")) ??
+      0;
+    return totalHectaresRestoredSum;
```
roguenet, line 78: This intermediate variable isn't doing anything useful - just return in the statement above.

apps/dashboard-service/src/dashboard/dto/total-serction-header.dto.ts
```diff
@@ -0,0 +1,33 @@
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { ApiProperty } from "@nestjs/swagger";
+import { populateDto } from "@terramatch-microservices/common/dto/json-api-attributes";
+
+@JsonApiDto({ type: "totalSectionHeaders" })
+export class TotalSectionHeaderDto {
```
roguenet, line 6: Please update the filename to match the class (`section` is misspelled in the filename)

apps/dashboard-service/src/dashboard/dto/total-serction-header.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+  totalHectaresRestoredGoal: number;
+
+  @ApiProperty()
+  totalTreesRestored: number;
+
+  @ApiProperty()
+  totalTreesRestoredGoal: number;
+
+  @ApiProperty()
+  lastUpdatedAt: string | null;
```
roguenet, line 32: This needs proper nullish definition:
```
@ApiProperty({ nullable: true, type: String })
lastUpdatedAt: string | null;
```

apps/dashboard-service/src/dashboard/total-section-header.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+import { DelayedJobDto } from "./delayed-job.dto";
+import { DelayedJob } from "@terramatch-microservices/database/entities";
+
+@Controller("dashboard/v3/totalSectionHeaders")
+export class TotalSectionHeaderController {
+  constructor(private readonly cacheService: CacheService) {}
+
+  @Get()
+  @JsonApiResponse([TotalSectionHeaderDto, DelayedJobDto])
+  @ApiOperation({ summary: "Get total section header" })
```
roguenet, line 17: Always provide an operationId - it will make the methods in the FE more sensible than the default `@ApiOperation({ operationId: "getTotalSectionHeaders", summary: "Get total section header" })`

apps/dashboard-service/src/dashboard/total-section-header.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+  async getTotalSectionHeader(@Query() query: DashboardQueryDto) {
+    const cacheKey = `dashboard:total-section-header|${this.cacheService.getCacheKeyFromQuery(query)}`;
+    const lastUpdatedAt = await this.cacheService.getTimestampForTotalSectionHeader(
+      this.cacheService.getCacheKeyFromQuery(query)
+    );
+    const cachedData = await this.cacheService.get(cacheKey);
+    const parseCachedData = typeof cachedData === "string" ? JSON.parse(cachedData) : cachedData;
```
roguenet, line 24: I think this kind of conversion would better handled in the cache service. The controller shouldn't need to know that sometimes a string comes out of the service.

apps/dashboard-service/src/dashboard/total-section-header.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+        totalTreesRestoredGoal,
+        lastUpdatedAt
+      })
+    );
+
+    document.addIndexData({
+      resource: "totalSectionHeaders",
+      requestPath: `/dashboard/v3/totalSectionHeaders${stableQuery}`,
+      ids: uuids
+    });
```
roguenet, line 66: This isn't an index endpoint, it's a single resource GET, so please drop the `addIndexData()` bit.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -13,9 +13,9 @@ import { ApiOperation } from "@nestjs/swagger";
 import { Op } from "sequelize";
 import { ExceptionResponse, JsonApiResponse } from "@terramatch-microservices/common/decorators";
 import { buildJsonApi, getDtoType, JsonApiDocument } from "@terramatch-microservices/common/util";
-import { DelayedJobDto } from "./dto/delayed-job.dto";
 import { DelayedJob } from "@terramatch-microservices/database/entities";
 import { DelayedJobBulkUpdateBodyDto } from "./dto/delayed-job-update.dto";
+import { DelayedJobDto } from "./dto/delayed-job.dto";
```
roguenet, line 18: This should be updated to pull from common, and the instance in this service should be removed.

apps/dashboard-service/src/dashboard/delayed-job.dto.ts
```diff
@@ -0,0 +1,76 @@
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { ApiProperty } from "@nestjs/swagger";
+import { populateDto } from "@terramatch-microservices/common/dto/json-api-attributes";
+import { DelayedJob } from "@terramatch-microservices/database/entities";
+
+const STATUSES = ["pending", "failed", "succeeded"];
+type Status = (typeof STATUSES)[number];
+
+@JsonApiDto({ type: "delayedJobs" })
+export class DelayedJobDto {
```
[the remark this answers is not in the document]
roguenet: I just saw that it was indeed moved to common, so this one should be removed.

libs/common/src/lib/util/json-api-builder.ts
```diff
@@ -210,3 +210,21 @@ export const getStableRequestQuery = (originalQuery: object) => {
...
+      const sorted = value.map(String).sort((a, b) => a.localeCompare(b));
+      const joined = sorted.join("+");
+      return `${encodeURIComponent(key)}=${encodeURIComponent(joined)}`;
+    }
+    return `${encodeURIComponent(key)}=${encodeURIComponent(String(value))}`;
+  });
+
+  const queryString = pairs.join("&");
+  return queryString.length === 0 ? "" : `?${queryString}`;
+};
```
roguenet, line 230: I haven't read through this carefully enough, but from the function name, it seems like it's doing the same thing as `getStableRequestQuery` (the method directly above this one). What's different about it?

LimberHope: I was actually using that function both in the backend and frontend to process the query and convert it into a string, since I use it as a key.
However, I ran into a special case with array-type parameters: even though the function in FE and BE is quite similar (based on getStableRequestQuery), the order of elements in the array when converted to a string ended up being different on each side.

This caused the key generated on the backend not to match the frontend key, which made the frontend request keep re-triggering endlessly because it saw the key as changing every time.

That’s why I implemented a similar but adjusted function to ensure consistent ordering and serialization of arrays between backend and frontend, to avoid those infinite loops.

apps/job-service/src/jobs/delayed-jobs.controller.ts
```diff
@@ -15,8 +15,8 @@ import { ExceptionResponse, JsonApiResponse } from "@terramatch-microservices/co
 import { buildJsonApi, getDtoType, JsonApiDocument } from "@terramatch-microservices/common/util";
 import { DelayedJob } from "@terramatch-microservices/database/entities";
 import { DelayedJobBulkUpdateBodyDto } from "./dto/delayed-job-update.dto";
-import { DelayedJobDto } from "./dto/delayed-job.dto";
 import { populateDto } from "@terramatch-microservices/common/dto/json-api-attributes";
+import { DelayedJobDto } from "../../../../libs/common/src/lib/dto/delayed-job.dto";
```
roguenet, line 19: Please use library-relative imports: 
```
import { DelayedJobDto } from "@terramatch-microservices/common/dto";
```
(you may need to add it to the index in that directory if you haven't already.

apps/dashboard-service/src/dashboard/dto/cache.service.ts
```diff
@@ -54,14 +54,18 @@ export class CacheService {
...
   getCacheParameterForProgrammes(programmes: string[]) {
-    if (programmes == null || programmes.length === 0) {
-      return "";
+    if (typeof programmes === "string") {
+      return (programmes = [programmes]);
     }
-    const sortedProgrammes = programmes.sort();
-    return sortedProgrammes.join(",");
+
+    return programmes.length === 0 ? "" : programmes.sort().join(",");
```
roguenet, line 61: A couple of things here:
* :art: You can leave as-is, but I generally avoid `typeof` and would use `if (!Array.isArray(programmes))` above
* There's no reason to assign to `programmes` when returning. That line can simply be `return [programmes];`
* Why is the if block above returning an array, but this line is intentionally converting an array to a string? 
* Same questions for the similar methods below.

apps/dashboard-service/src/dashboard/dto/cache.service.ts
```diff
@@ -54,18 +54,10 @@ export class CacheService {
   }
 
   getCacheParameterForProgrammes(programmes: string[]) {
-    if (typeof programmes === "string") {
-      return (programmes = [programmes]);
-    }
-
     return programmes.length === 0 ? "" : programmes.sort().join(",");
```
roguenet, line 57: `getCacheParameterForOrganisationType` needs this update as well.

PR #164, opened 2025-05-09
[TM-1775] implement file upload media update

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-1775)

### DIFF
--- a/apps/unified-database-service/src/airtable/airtable.processor.ts
+++ b/apps/unified-database-service/src/airtable/airtable.processor.ts
@@ -7,6 +7,7 @@ import {
   ApplicationEntity,
   DemographicEntity,
   DemographicEntryEntity,
+  FinancialIndicatorEntity,
   FundingProgrammeEntity,
   NurseryEntity,
   NurseryReportEntity,
@@ -27,6 +28,7 @@ export const AIRTABLE_ENTITIES = {
   applications: ApplicationEntity,
   demographics: DemographicEntity,
   demographicEntries: DemographicEntryEntity,
+  financialIndicators: FinancialIndicatorEntity,
   fundingProgrammes: FundingProgrammeEntity,
   nurseries: NurseryEntity,
   nurseryReports: NurseryReportEntity,
@@ -77,6 +79,7 @@ export class AirtableProcessor extends WorkerHost {
     const apiKey = this.config.get<string>("AIRTABLE_API_KEY");
     const baseId = this.config.get<string>("AIRTABLE_BASE_ID");
     if (apiKey == null || baseId == null) {
+      /* istanbul ignore next */
       throw new InternalServerErrorException("Airtable API key and base ID must be set");
     }
 
@@ -101,6 +104,7 @@ export class AirtableProcessor extends WorkerHost {
     }
   }
 
+  /* istanbul ignore next */
   @OnWorkerEvent("failed")
   async onFailed(job: Job, error: Error) {
     Sentry.captureException(error);
@@ -155,7 +159,7 @@ export class AirtableProcessor extends WorkerHost {
 
     try {
       await this.slack.sendTextToChannel(`[${process.env.DEPLOY_ENV}]: ${message}`, channel);
-    } catch (error) {
+    } catch (error) /* istanbul ignore next */ {
       // Don't allow a failure in slack sending to hose our process, but do log it and send it to Sentry
       Sentry.captureException(error);
       this.logger.error("Send to slack failed", error.stack);
--- a/apps/unified-database-service/src/airtable/entities/airtable-entity.spec.ts
+++ b/apps/unified-database-service/src/airtable/entities/airtable-entity.spec.ts
@@ -4,6 +4,7 @@ import {
   Application,
   Demographic,
   DemographicEntry,
+  FinancialIndicator,
   Framework,
   FundingProgramme,
   Nursery,
@@ -20,6 +21,7 @@ import {
   ApplicationFactory,
   DemographicEntryFactory,
   DemographicFactory,
+  FinancialIndicatorFactory,
   FormSubmissionFactory,
   FundingProgrammeFactory,
   NurseryFactory,
@@ -40,6 +42,7 @@ import {
   ApplicationEntity,
   DemographicEntity,
   DemographicEntryEntity,
+  FinancialIndicatorEntity,
   FundingProgrammeEntity,
   NurseryEntity,
   NurseryReportEntity,
@@ -51,7 +54,7 @@ import {
   SiteReportEntity,
   TreeSpeciesEntity
 } from "./";
-import { orderBy, sortBy } from "lodash";
+import { orderBy, sortBy, uniq } from "lodash";
 import { Model } from "sequelize-typescript";
 import { FrameworkKey } from "@terramatch-microservices/database/constants/framework";
 import { FindOptions, Op } from "sequelize";
@@ -498,6 +501,49 @@ describe("AirtableEntity", () => {
           dueAt
         }
       }));
+    });
+  });
+
+  describe("FinancialIndicatorEntity", () => {
+    let indicators: FinancialIndicator[];
+    let organisationUuids: Record<number, string>;
+
+    beforeAll(async () => {
+      await FinancialIndicator.truncate();
+
+      const allIndicators = await FinancialIndicatorFactory.createMany(10);
+      await allIndicators[2].destroy();
+      await allIndicators[7].destroy();
+      allIndicators.push(
+        await FinancialIndicatorFactory.create({ organisationId: allIndicators[0].organisationId }),
+        await FinancialIndicatorFactory.create({ organisationId: allIndicators[2].organisationId })
+      );
+
+      indicators = allIndicators.filter(indicator => !indicator.isSoftDeleted());
+
+      organisationUuids = {};
+      for (const { id, uuid } of await Organisation.findAll({
+        where: { id: uniq(indicators.map(({ organisationId }) => organisationId)) },
+        attributes: ["id", "uuid"]
+      })) {
+        organisationUuids[id] = uuid;
+      }
+    });
+
+    it("sends all records to airtable", async () => {
+      await testAirtableUpdates(
+        new FinancialIndicatorEntity(dataApi),
+        indicators,
+        ({ uuid, collection, organisationId, amount, description }) => ({
+          fields: {
+            uuid,
+            collection,
+            amount,
+            description,
+            organisationUuid: organisationUuids[organisationId]
+          }
+        })
+      );
     });
   });
 
--- a/apps/unified-database-service/src/airtable/entities/financial-indicator.airtable-entity.ts
+++ b/apps/unified-database-service/src/airtable/entities/financial-indicator.airtable-entity.ts
@@ -1,0 +1,39 @@
+import { FinancialIndicator, Organisation } from "@terramatch-microservices/database/entities";
+import { AirtableEntity, associatedValueColumn, ColumnMapping } from "./airtable-entity";
+import { uniq } from "lodash";
+
+type FinancialIndicatorAssociations = {
+  organisationUuid?: string;
+};
+
+const COLUMNS: ColumnMapping<FinancialIndicator, FinancialIndicatorAssociations>[] = [
+  "uuid",
+  "createdAt",
+  "updatedAt",
+  "collection",
+  "amount",
+  "description",
+  associatedValueColumn("organisationUuid", "organisationId")
+];
+
+export class FinancialIndicatorEntity extends AirtableEntity<FinancialIndicator, FinancialIndicatorAssociations> {
+  readonly TABLE_NAME = "Financial Indicators";
+  readonly COLUMNS = COLUMNS;
+  readonly MODEL = FinancialIndicator;
+
+  protected async loadAssociations(indicators: FinancialIndicator[]) {
+    const orgs = await Organisation.findAll({
+      where: { id: uniq(indicators.map(({ organisationId }) => organisationId)) },
+      attributes: ["id", "uuid"]
+    });
+    return indicators.reduce(
+      (associations, { id, organisationId }) => ({
+        ...associations,
+        [id]: {
+          organisationUuid: orgs.find(({ id }) => id === organisationId)?.uuid
+        }
+      }),
+      {} as Record<number, FinancialIndicatorAssociations>
+    );
+  }
+}
--- a/apps/unified-database-service/src/airtable/entities/index.ts
+++ b/apps/unified-database-service/src/airtable/entities/index.ts
@@ -1,6 +1,7 @@
 export * from "./application.airtable-entity";
 export * from "./demographic.airtable-entity";
 export * from "./demographic-entry.airtable-entity";
+export * from "./financial-indicator.airtable-entity";
 export * from "./funding-programme.airtable-entity";
 export * from "./nursery.airtable-entity";
 export * from "./nursery-report.airtable-entity";
--- a/libs/database/src/lib/entities/financial-indicator.entity.ts
+++ b/libs/database/src/lib/entities/financial-indicator.entity.ts
@@ -1,5 +1,5 @@
-import { AutoIncrement, Column, Index, Model, PrimaryKey, Table } from "sequelize-typescript";
-import { BIGINT, STRING, UUID, UUIDV4 } from "sequelize";
+import { AllowNull, AutoIncrement, Column, Index, Model, PrimaryKey, Table } from "sequelize-typescript";
+import { BIGINT, DECIMAL, STRING, TEXT, UUID, UUIDV4 } from "sequelize";
 
 @Table({ tableName: "financial_indicators", underscored: true, paranoid: true })
 export class FinancialIndicator extends Model<FinancialIndicator> {
@@ -24,5 +24,10 @@ export class FinancialIndicator extends Model<FinancialIndicator> {
   @Column(STRING)
   collection: string;
 
-  // TODO: complete remaining fields
+  @AllowNull
+  @Column(DECIMAL(15, 2))
+  amount: number | null;
+
+  @Column(TEXT)
+  description: string | null;
 }
--- a/libs/database/src/lib/factories/financial-indicator.factory.ts
+++ b/libs/database/src/lib/factories/financial-indicator.factory.ts
@@ -1,9 +1,11 @@
-import { FinancialIndicator } from "../entities";
 import { FactoryGirl } from "factory-girl-ts";
+import { FinancialIndicator } from "../entities";
+import { OrganisationFactory } from ".";
 import { faker } from "@faker-js/faker";
-import { OrganisationFactory } from "./organisation.factory";
 
 export const FinancialIndicatorFactory = FactoryGirl.define(FinancialIndicator, async () => ({
   organisationId: OrganisationFactory.associate("id"),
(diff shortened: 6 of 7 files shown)


pachonjcl:
Hi @roguenet before I move forward I would like to have an early review about the current structure.
If there is anything that you would like to address, please mention it.

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -0,0 +1,30 @@
+import { Controller, Post, Param, NotFoundException, Req } from "@nestjs/common";
+import { FileUploadService } from "../trees/file-upload.service";
+import { EntitiesService } from "./entities.service";
+import { EntityModel } from "@terramatch-microservices/database/constants/entities";
+import { SpecificEntityDto } from "./dto/specific-entity.dto";
+import { PolicyService } from "@terramatch-microservices/common/policies/policy.service";
+import { FastifyRequest } from "fastify";
+
+@Controller("entities/v3/file/upload")
```
roguenet, line 9: Two things:
* Please use plural for the base path name
* `upload` is redundant. If you have an API endpoint that accepts a post with media on it, it's pretty clear that's an upload.

So, I'd use `@Controller("entities/v3/files")` for this. In the `app.module.ts`, you'll need to move this new controller ahead of `EntitiesController` in the list of controllers to avoid getting swallowed up by the wildcard routes in that controller (see my in-progress tasks branch: https://github.com/wri/terramatch-microservices/blob/dc8e2a82c1c5ad93717cda01aa70ce4e4900ff78/apps/entity-service/src/app.module.ts#L15-L18)

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(
+    private readonly fileUploadService: FileUploadService,
+    private readonly entitiesService: EntitiesService,
+    private readonly policyService: PolicyService
+  ) {}
+
+  @Post(":collection/:entity/:uuid")
+  async uploadFile<T extends EntityModel>(
+    @Param("collection") collection: string,
+    @Param() { entity, uuid }: SpecificEntityDto,
```
roguenet, line 20: I kind of dislike using the `@Param("name")` construct because it doesn't allow for doing validation (not super important here, but as a pattern I think we should avoid it). Also, I don't think there's any real reason to go out of the way to re-use DTOs from the base entities service. I'd create a param DTO specific to this endpoint and use it here to collect all three. It could extend the `SpecificEntityDto` to avoid some duplication if you like.

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -0,0 +1,30 @@
+import { Controller, Post, Param, NotFoundException, Req } from "@nestjs/common";
+import { FileUploadService } from "../trees/file-upload.service";
```
roguenet, line 2: Why is this under `trees`?

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+    private readonly policyService: PolicyService
+  ) {}
+
+  @Post(":collection/:entity/:uuid")
+  async uploadFile<T extends EntityModel>(
+    @Param("collection") collection: string,
+    @Param() { entity, uuid }: SpecificEntityDto,
+    @Req() req: FastifyRequest
+  ) {
+    const file = await (req as any).file();
```
roguenet, line 23: Does this really need `any`? Let's try to avoid that. Here, and elsewhere in this PR that uses `as any`

apps/entity-service/src/trees/file-upload.service.ts
```diff
@@ -0,0 +1,113 @@
...
+  private validateFile(file: any, configuration: any) {
+    const validationFileTypes = VALIDATION.VALIDATION_FILE_TYPES[configuration.validation];
+    const validationRules = VALIDATION.VALIDATION_RULES[validationFileTypes];
+
+    const validations = validationRules.split("|");
+
+    const mimeTypeValidation = validations.find(validation => validation.startsWith("mimes:"));
+    const sizeValidation = validations.find(validation => validation.startsWith("size:"));
+
+    if (mimeTypeValidation) {
```
roguenet, line 94: Remember to use explicit null checks: `if (mimeTypeValidation != null)`

apps/entity-service/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "files": [],
-  "include": [],
+  "include": ["src/**/*"],
```
roguenet, line 4: What is this change needed for?

instrument-sentry.ts
```diff
@@ -1,11 +1,12 @@
 import * as Sentry from "@sentry/nestjs";
 import { nodeProfilingIntegration } from "@sentry/profiling-node";
+import { Integration } from "@sentry/core";
 
 if (process.env.SENTRY_DSN != null) {
   Sentry.init({
     dsn: process.env.SENTRY_DSN,
     environment: process.env.DEPLOY_ENV,
-    integrations: [nodeProfilingIntegration()],
+    integrations: [nodeProfilingIntegration() as unknown as Integration],
```
roguenet, line 9: I'm surprised by this change as well; why is it needed?

libs/common/src/lib/media/media.service.ts
```diff
@@ -1,10 +1,39 @@
...
+    const command = new PutObjectCommand({
+      Bucket: bucket,
+      Key: key,
+      Body: buffer,
+      ContentType: contentType,
+      ACL: "public-read"
+    });
+
+    await this.s3.send(command);
+    console.log(`Uploaded ${key}`);
```
roguenet, line 35: We'll want to make use of TMLogger instead of console.log in the final code. This one seems like a reasonable thing to promote to a real log.

roguenet (commented):
I think this is all heading in a good direction. I tried not to go too deep on the review, but called out a few things that stood out to me.

apps/entity-service/src/entities/file-upload.controller.ts
pachonjcl: changed

apps/entity-service/src/entities/file-upload.controller.ts
pachonjcl: moved

apps/entity-service/tsconfig.json
pachonjcl: removed

apps/entity-service/src/entities/file-upload.controller.ts
pachonjcl: removed

apps/entity-service/src/trees/file-upload.service.ts
pachonjcl: got it.

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Controller("entities/v3/files")
+export class FileUploadController {
+  constructor(
+    private readonly fileUploadService: FileUploadService,
+    private readonly entitiesService: EntitiesService,
+    private readonly policyService: PolicyService
+  ) {}
+
+  @Post("/:collection/:entity/:uuid")
```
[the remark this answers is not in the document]
pachonjcl: changed

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -0,0 +1,28 @@
...
+  async uploadFile<T extends EntityModel>(
```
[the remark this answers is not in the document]
pachonjcl: added

apps/entity-service/src/file/file-upload.service.ts
```diff
@@ -0,0 +1,189 @@
...
+      file: null,
+      isPublic: false,
+      lat: 0,
+      lng: 0
+    };
+
+    for await (const part of parts) {
+      if ((part as Part).file != null) {
+        // @ts-ignore
+        extraFields.file = part;
```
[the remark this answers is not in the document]
pachonjcl: removed

apps/entity-service/src/main.ts
```diff
@@ -6,13 +6,20 @@ import { NestFactory } from "@nestjs/core";
...
+  const adapter = new FastifyAdapter();
+
+  await adapter.register(multipart);
+
+  const app = await NestFactory.create<NestFastifyApplication>(AppModule, adapter, {
     logger: new TMLogger()
   });
-  app.set("query parser", "extended");
+  // TODO: register back this query parser
+  // app.set("query parser", "extended");
```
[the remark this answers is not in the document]
pachonjcl: addressed

package.json
```diff
@@ -74,6 +79,7 @@
     "@swc-node/register": "~1.10.10",
     "@swc/core": "~1.11.11",
     "@swc/helpers": "~0.5.15",
+    "@types/busboy": "^1.5.4",
```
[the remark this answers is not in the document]
pachonjcl: I have removed

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -1,28 +1,71 @@
...
+    operationId: "uploadFile",
+    summary: "Upload a file to a media collection",
+    description: "Upload a file to a media collection"
+  })
+  @ExceptionResponse(UnauthorizedException, {
+    description: "Authentication failed, or resource unavailable to current user."
+  })
+  @ExceptionResponse(NotFoundException, { description: "Resource not found." })
+  @ExceptionResponse(BadRequestException, { description: "Invalid request." })
+  @ExceptionResponse(InternalServerErrorException, { description: "Internal server error." })
```
roguenet, line 47: I don't think it's useful to document InternalServerErrorException. That particular exception is really meant for something that's broken about the internal state of the server (probably a misconfiguration). It's not actionable on the client end because it isn't the result of bad input, and we would really never expect it to happen, especially in prod.

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -1,28 +1,71 @@
...
+  @UseInterceptors(FileInterceptor("file"))
```
roguenet, line 48: This endpoint still appears to need a `@JsonApiResponse` decorator.

apps/entity-service/src/file/file-upload.service.ts
```diff
@@ -98,42 +109,41 @@ export class FileUploadService {
       createdBy: this.entitiesService.userId
     };
 
-    const dbMedia = await Media.create(media as Media);
+    console.log(media);
```
roguenet, line 112: Please make sure to remove any use of `console.log` before this gets merged.

apps/entity-service/src/file/file-upload.service.ts
```diff
@@ -98,42 +109,41 @@ export class FileUploadService {
...
 
-    return new MediaDto(dbMedia, {
-      url: this.mediaService.getUrl(dbMedia),
-      thumbUrl: this.mediaService.getUrl(dbMedia, "thumbnail"),
-      entityType: entity,
-      entityUuid: model.uuid
-    });
+    const dbMedia = new Media(media as Media);
+    return dbMedia.save();
```
roguenet, line 115: In my previous feedback on this, I was meaning that up above when you create `media`, you could do instead as:
```
const media = new Media();
media.collectionName = collection;
... etc

await media.save();
```

I don't think it's useful to simply replace a single line `await Media.create()` with a two line new / save.

apps/entity-service/src/file/file-upload.service.ts
```diff
@@ -98,42 +109,41 @@ export class FileUploadService {
...
   }
 
-  private getConfiguration(entity: EntityClass<EntityModel>, collection: string): MediaConfiguration {
-    const configuration = (entity as any).MEDIA[collection];
+  private getConfiguration(entityModel: EntityClass<EntityModel>, collection: string): MediaConfiguration {
+    // TODO: ask Nathan what is the best way to check if a EntityClass has a MEDIA property and how to type it
```
roguenet, line 119: For this, I would accept a specific type of model in the method declaration. See `types/util.ts` in the DB lib for an example of this pattern, and find their usages. I'm thinking something specifically similar to `LaravelModelCtor`.

instrument-sentry.ts
```diff
@@ -1,11 +1,12 @@
 import * as Sentry from "@sentry/nestjs";
 import { nodeProfilingIntegration } from "@sentry/profiling-node";
+import { Integration } from "@sentry/core";
 
 if (process.env.SENTRY_DSN != null) {
   Sentry.init({
     dsn: process.env.SENTRY_DSN,
     environment: process.env.DEPLOY_ENV,
-    integrations: [nodeProfilingIntegration()],
+    integrations: [nodeProfilingIntegration() as Integration],
```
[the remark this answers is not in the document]
roguenet: Odd... this file has been here for months with no problems. Any idea why it would be just starting now? My local environment on the `staging` branch shows no type errors on this line, and my IDE shows that the return value from `nodeProfilingIntegration` is properly typed: 
image

apps/entity-service/src/file/file-upload.service.ts
```diff
@@ -51,11 +60,15 @@ const VALIDATION = {
...
-
-  public async uploadFile(model: EntityModel, entity: EntityType, collection: string, req: FastifyRequest) {
-    const { file, ...extraFields } = await this.extractRequestData(req);
+  constructor(private readonly mediaService: MediaService, private readonly entitiesService: EntitiesService) {}
 
+  public async uploadFile(
+    model: EntityModel,
+    entity: EntityType,
+    collection: string,
+    file: Express.Multer.File,
```
roguenet, line 69: This, combined with the naked `import "multer";` above is a bit weird. Is it possible to something more like `import { File } from "multer";` and avoid accessing the type from the global express namespace? Same question for the controller.

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -1,28 +1,71 @@
...
+import { ExceptionResponse } from "@terramatch-microservices/common/decorators/exception-response.decorator";
+import { ApiExtraModels, ApiOperation } from "@nestjs/swagger";
+import { FileInterceptor } from "@nestjs/platform-express";
+import "multer";
+import { buildJsonApi } from "@terramatch-microservices/common/util/json-api-builder";
+import { MediaDto } from "./dto/media.dto";
+import { MediaService } from "@terramatch-microservices/common/media/media.service";
 
 @Controller("entities/v3/files")
+@ApiExtraModels(MediaDto)
```
roguenet, line 27: I believe this won't be needed once the correct `@JsonApiResponse` decorator gets added to the endpoint.

apps/entity-service/src/file/file-upload.service.ts
pachonjcl: there are some problems with the typing otherwise, I will give another look, but for now this is the way it works without type errors

instrument-sentry.ts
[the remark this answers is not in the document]
roguenet: Can you send me a link? It doesn't make any sense for this thing that is running fine on [other active branches](https://github.com/wri/terramatch-microservices/actions/runs/15545083093/job/43766714847?pr=196) to have a problem here.

apps/entity-service/src/entities/file-upload.controller.ts
```diff
@@ -44,16 +41,15 @@ export class FileUploadController {
   })
   @ExceptionResponse(NotFoundException, { description: "Resource not found." })
   @ExceptionResponse(BadRequestException, { description: "Invalid request." })
-  @ExceptionResponse(InternalServerErrorException, { description: "Internal server error." })
-  @UseInterceptors(FileInterceptor("file"))
-  async uploadFile<T extends EntityModel>(
+  @UseInterceptors(FileInterceptor("uploadFile"))
+  @JsonApiResponse({ data: MediaDto })
```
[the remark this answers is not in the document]
pachonjcl: changed

libs/database/src/lib/entities/form.entity.ts
```diff
@@ -0,0 +1,77 @@
+import { AllowNull, AutoIncrement, Column, ForeignKey, Index, Model, PrimaryKey, Table } from "sequelize-typescript";
+import { BIGINT, BOOLEAN, DATE, STRING, TINYINT, UUID, UUIDV4 } from "sequelize";
+import { FrameworkKey } from "../constants/framework";
+import { User } from "./user.entity";
+import { MediaConfiguration } from "../constants/media-owners";
+
+@Table({ tableName: "v2_forms", underscored: true, paranoid: true })
```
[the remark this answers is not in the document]
pachonjcl: sorry about that. I let the autocomplete go crazy

instrument-sentry.ts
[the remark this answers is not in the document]
roguenet: Ok, this probably isn't related to your env, but was a problem in the package.json. Somewhere along the way over upgrades and various package installs, the version for `@sentry/nestjs` lost the `^`, but the profiling extension still had it, so this problem was likely to come up sooner or later: https://github.com/wri/terramatch-microservices/pull/164/commits/eb0b6ac2fbcdb789c8e55269c69ef719f299d99f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R31

libs/common/src/lib/policies/form.policy.ts
```diff
@@ -15,7 +16,7 @@ export class FormPolicy extends UserPermissionsPolicy {
 
     return (this._user = await User.findOne({
       where: { id: this.userId },
-      attributes: ["id"]
+      attributes: ["uuid"]
```
roguenet, line 19: This `getUser()` method is no longer used and can simply be removed.

libs/common/src/lib/policies/form-question-option.policy.ts
```diff
@@ -1,19 +1,37 @@
...
 
-      this.builder.can(["read", "update", "uploadFiles"], FormQuestionOption, {
-        formQuestionId: { $in: formIds }
-      });
+      const formQuestionsIds = formQuestion.map(q => q.id);
+      if (formQuestionsIds.length > 0) {
+        this.builder.can(["uploadFiles"], FormQuestionOption, {
+          formQuestionId: { $in: formQuestionsIds }
+        });
+      }
```
roguenet, line 34: The `getUser()` method in this policy may be removed now as well.

roguenet (commented):
Getting close! See my remaining comment from the previous review on the impact story policy; there's an open question there.

libs/common/src/lib/policies/form-question-option.policy.ts
pachonjcl: removed

libs/common/src/lib/policies/impact-story.policy.ts
```diff
@@ -3,7 +3,7 @@ import { ImpactStory } from "@terramatch-microservices/database/entities";
 
 export class ImpactStoryPolicy extends UserPermissionsPolicy {
   async addRules() {
-    if (this.permissions.includes("impact-stories-manage")) {
+    if (this.frameworks.length > 0 || this.permissions.includes("impact-stories-manage")) {
```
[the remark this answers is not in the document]
pachonjcl: ok

roguenet (approved):

PR #168, opened 2025-05-19
[TM-1995] Fix Remove Unnecesary column

Scriptmatico:


roguenet (commented):
This is not the right fix. Please see https://github.com/wri/terramatch-microservices/pull/169 and close this PR

PR #170, opened 2025-05-19
[TM-2040] boundingbox resource

egrojMonroy:
This bbox resource returns the bbox of all provided query param:
project, polygon, site, country or landscape.


apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+export class BoundingBoxController {
+  constructor(private readonly boundingBoxService: BoundingBoxService) {}
+
+  @Get()
+  @ApiOperation({
+    operationId: "boundingBoxGet",
+    summary: "Get a bounding box for a polygon, site, project, or country/landscape"
+  })
+  @JsonApiResponse(BoundingBoxDto)
+  async getBoundingBox(@Query() query: BoundingBoxQueryDto): Promise<JsonApiDocument> {
```
roguenet, line 20: This should have `@ExceptionResponse`s to declare which error codes are possible. I see `BadRequestException` below. What I see missing is any `PolicyService` authorize calls. Is this truly a globally accessible resource? I would think this should only be available to the folks that are able to read the project or site that is being bounded.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+    const providedParams: string[] = [];
+
+    if (query.polygonUuid !== undefined && query.polygonUuid !== null && query.polygonUuid !== "") {
```
roguenet, line 23: Using `!== undefined && ... !== null` is not necessary. Please use `!= null` to cover both cases. And in this case, if you wanted to you could use the lodash `isEmpty` method to make this simply `if (!isEmpty(query.polygonUuid) {`

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (query.projectUuid !== undefined && query.projectUuid !== null && query.projectUuid !== "") {
+      providedParams.push("projectUuid");
+    }
+
+    const hasCountry = query.country !== undefined && query.country !== null && query.country !== "";
+    const hasLandscapes =
+      query.landscapes !== undefined &&
+      query.landscapes !== null &&
+      Array.isArray(query.landscapes) &&
+      query.landscapes.length > 0;
```
roguenet, line 40: This can be simply `const hasLandscapes = (query.landscapes?.length ?? 0) > 0;`, or `const hasLandscapes = query.landscapes != null && query.landscapes.length > 0;`. The Query DTO validation will complain if a value is provided that is not an array, so the `Array.isArray` check is definitely superfluous.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+      providedParams.push("country/landscapes");
+    }
+
+    if (providedParams.length > 1) {
+      throw new BadRequestException(
+        `Mutually exclusive parameters provided: ${providedParams.join(", ")}. Please provide only one entity type.`
+      );
+    }
+
+    let result: BoundingBoxDto;
```
roguenet, line 52: I try to avoid mutability when possible. This `result` variable is never used once it gets assigned in a given block, so there's no reason to declare it in this scope. Use `const result = ` in each of the `if` blocks below instead.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    if (query.polygonUuid !== undefined && query.polygonUuid !== null && query.polygonUuid !== "") {
```
roguenet, line 54: These checks have already been done when assigning to `providedParams`. This block can then instead be `if (providedParams[0] === "polygonUuid") {`. If you want to be even more rigorous, you could do either or both of these:
* Provide a type of the possible values of `providedParams` and use it in the declaration `const providedParams: ProvidedParams[] = []`
* Make this a switch because there should be only one by the time we get to this point in the controller (see comment below about moving the missing params `BadRequestException` up above this point.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (hasCountry || hasLandscapes) {
+      const landscapes: string[] = hasLandscapes && Array.isArray(query.landscapes) ? query.landscapes : [];
+      const country = hasCountry ? query.country : "global";
+
+      result = await this.boundingBoxService.getCountryLandscapeBoundingBox(country as string, landscapes);
+      const id = `${country}-${landscapes.join("-")}`;
+      return buildJsonApi(BoundingBoxDto).addData(id, result).document.serialize();
+    }
+
+    throw new BadRequestException(
```
roguenet, line 78: This can be thrown before any of these `if` blocks by checking after the `providedParams.length > 1` for `providedParams.length === 0`. Always prefer throwing validation errors as early as possible to avoid unnecessary logic blocks.

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+  maxLng: number;
+  maxLat: number;
+}
+
+@Injectable()
+export class BoundingBoxService {
+  constructor(private readonly dataApiService: DataApiService, private readonly configService: ConfigService) {}
+
+  private createBoundingBoxDto(minLng: number, minLat: number, maxLng: number, maxLat: number): BoundingBoxDto {
+    return new BoundingBoxDto({
```
roguenet, line 34: :art: Up to you, but since this is the only place that this DTO gets created, and this is the place that accepts the correct numbers instead of a number array, I would change this method body to:
```
return populateDto(BoundingBoxDto, { bbox: [minLng, minLat, maxLng, maxLat] });
```
and remove the constructor from `BoundingBoxDto`.

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+        [Sequelize.fn("ST_ASGEOJSON", Sequelize.fn("ST_Envelope", Sequelize.col(geometryColumn))), "envelope"]
+      ]
+    });
+
+    if (envelopes.length === 0) {
+      const errorMsg =
+        identifier !== undefined && identifier !== null && identifier !== ""
+          ? `No ${entityType.toLowerCase()} found with UUID ${identifier}`
+          : `No ${entityType.toLowerCase()} found with the provided criteria`;
+      throw new NotFoundException(errorMsg);
```
roguenet, line 80: Make sure to include this in the exception responses on the controller.

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+    );
+  }
+
+  async getSiteBoundingBox(siteUuid: string): Promise<BoundingBoxDto> {
+    const site = await Site.findOne({
+      where: { uuid: siteUuid },
+      attributes: ["uuid"]
+    });
+
+    if (site === null || site === undefined) {
```
roguenet, line 103: Same thing here as in the cases in the controller; this should simply be `if (site == null) {` (and please check this service for any additional cases like this)

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+  async getProjectBoundingBox(projectUuid: string): Promise<BoundingBoxDto> {
+    const project = await Project.findOne({
+      where: { uuid: projectUuid },
+      attributes: ["id"]
+    });
+
+    if (project === null || project === undefined) {
+      throw new NotFoundException(`${EntityType.PROJECT} with UUID ${projectUuid} not found`);
+    }
+
```
roguenet, line 140: As mentioned in the controller, there should be policy checks on this stuff. In these methods is where I would do it with `await this.policyService.authorize('read', project)` (and in the other entity specific methods in this service)

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+    return this.getBoundingBoxFromGeometries(
+      PolygonGeometry,
+      { uuid: { [Op.in]: polygonUuids } },
+      "geom",
+      EntityType.POLYGON
+    );
+  }
+
+  async getCountryLandscapeBoundingBox(country: string, landscapes: string[]): Promise<BoundingBoxDto> {
```
roguenet, line 174: This one I think can ignore the policy service.

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+        maxLat = Math.max(maxLat, north);
+        minLat = Math.min(minLat, south);
+        hasBounds = true;
+      } catch (error) {
+        if (!(error instanceof NotFoundException)) {
+          throw error;
+        }
+      }
+    }
+    if (typeof country === "string" && country !== "") {
```
roguenet, line 204: I think this typeof call is unnecessary. Between the validation in the query DTO and the TS compiler doing it's thing, we can be pretty certain a non-string didn't sneak in here.

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+
+        try {
+          const envelopeData = await this.dataApiService.getCountryEnvelope(countryIso);
+
+          if (
+            envelopeData.length > 0 &&
+            typeof envelopeData[0] === "object" &&
+            envelopeData[0] !== null &&
+            "envelope" in envelopeData[0] &&
+            typeof envelopeData[0].envelope === "string"
```
roguenet, line 216: These typeof calls would be a little cleaner with lodash's `isString` and `isObject` methods

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -0,0 +1,271 @@
...
+                      minLng = Math.min(minLng, lng);
+                      maxLat = Math.max(maxLat, lat);
+                      minLat = Math.min(minLat, lat);
+                    }
+                  }
+                }
+                hasBounds = true;
+              }
+            } catch (jsonParseError) {
+              console.error("Error parsing envelope JSON:", jsonParseError);
```
roguenet, line 250: There should be no `console` calls in this repo. Please use the `TMLogger`. Lots of examples of how to get one and use it in other services.

apps/entity-service/src/bounding-boxes/dto/bounding-box-query.dto.ts
```diff
@@ -0,0 +1,57 @@
...
+  projectUuid?: string;
+
+  @ApiPropertyOptional({
+    description: "Array of project UUIDs to get the combined bounding box of their centroids",
+    type: [String]
+  })
+  @IsOptional()
+  @IsArray()
+  @IsUUID(undefined, { each: true })
+  @Transform(({ value }) => (typeof value === "string" ? [value] : value))
```
roguenet, line 37: I would prefer not to make it possible to send a string value to this param. It should always be specified as `projectUuids[]=` in the query, which makes this transform unnecessary.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: changed

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: changed

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: changed

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
egrojMonroy: changed with lodash

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
egrojMonroy: using methods

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
egrojMonroy: changed

apps/entity-service/src/bounding-boxes/dto/bounding-box-query.dto.ts
egrojMonroy: changed. And also projectUuids was dead code, because the bbox for multiple projects is not used anymore as an endpoint because it will come from the lat - long from the table itself.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: added.
Also the policy is added for each case. 
In the case of polygon, I needed to update the sitepolygon policy.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: added

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: added

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
egrojMonroy: changed

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
egrojMonroy: added in controller

apps/entity-service/src/bounding-boxes/bounding-box.service.ts
```diff
@@ -85,6 +93,38 @@ export class BoundingBoxService {
...
+      attributes: ["uuid"]
+    });
+
+    if (polygon === null) {
+      throw new NotFoundException(`${EntityType.POLYGON} with UUID ${polygonUuid} not found`);
+    }
+
+    const sitePolygon = await SitePolygon.findOne({
+      where: {
+        polygonUuid,
```
roguenet, line 107: Let's avoid the intermediate polygon query with a subquery:
```
polygonUuid: Subquery.select(PolygonGeometry, "uuid").eq(uuid, polygonUuid).literal,
```

egrojMonroy: I just realized I forgot to delete taht code from service. 
But this subquery is implemented in controller, where is still using this query.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -68,27 +70,22 @@ export class BoundingBoxController {
...
-              association: "site",
-              required: true
+        const site = await Site.findOne({
+          where: {
+            uuid: {
+              [Op.in]: Subquery.select(SitePolygon, "siteUuid").eq("polygonUuid", polygonUuid).literal
             }
-          ],
-          attributes: ["id", "uuid", "polygonUuid"]
+          }
```
roguenet, line 78: Since this is just pulling the site for the policy, let's set the attributes to `["frameworkKey", "projectId"]`, which appears to be all that's needed by the site policy.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
```diff
@@ -133,8 +130,8 @@ export class BoundingBoxController {
       }
 
       case "country/landscapes": {
-        const landscapes: string[] = hasLandscapes && Array.isArray(query.landscapes) ? query.landscapes : [];
         const country = hasCountry ? (query.country as string) : "global";
+        const landscapes: string[] = hasLandscapes ? (query.landscapes as string[]) : [];
```
roguenet, line 134: `query.landscapes` already has the right typing, so this `as` should not be needed. Same with `query.country` above.

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: ok

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
egrojMonroy: right, i removed al unnecesary typing of the entire file

apps/entity-service/src/bounding-boxes/bounding-box.controller.ts
roguenet: This still needs addressing.

roguenet:
Code looks good, but I see that there's an error in the bounding box specs.

roguenet (dismissed):

PR #188, opened 2025-06-02
[TM-2083] tree restoration goals

egrojMonroy:
This gets the data for tree restoration goals. 
Here there is no need for delayedjob instance because the largest query now takes up to 3 secs. 
So it only has a cache.

### DIFF
--- a/apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.dto.ts
+++ b/apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.dto.ts
@@ -1,0 +1,49 @@
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { ApiProperty } from "@nestjs/swagger";
+import { populateDto } from "@terramatch-microservices/common/dto/json-api-attributes";
+
+@JsonApiDto({ type: "treeRestorationGoals" })
+export class TreeRestorationGoalDto {
+  constructor(data: TreeRestorationGoalDto) {
+    populateDto<TreeRestorationGoalDto>(this, data);
+  }
+
+  @ApiProperty({
+    description: "Total number of trees grown goal for for-profit organizations",
+    type: Number
+  })
+  forProfitTreeCount: number;
+
+  @ApiProperty({
+    description: "Total number of trees grown goal for non-profit organizations",
+    type: Number
+  })
+  nonProfitTreeCount: number;
+
+  @ApiProperty({
+    description: "Total trees grown goal across all organizations",
+    type: Number
+  })
+  totalTreesGrownGoal: number;
+
+  @ApiProperty({
+    description: "Total number of trees under restoration across all organizations",
+    type: Number
+  })
+  treesUnderRestorationActualTotal: number;
+
+  @ApiProperty({
+    description: "Total number of trees under restoration for for-profit organizations",
+    type: Number
+  })
+  treesUnderRestorationActualForProfit: number;
+
+  @ApiProperty({
+    description: "Total number of trees under restoration for non-profit organizations",
+    type: Number
+  })
+  treesUnderRestorationActualNonProfit: number;
+
+  @ApiProperty({ nullable: true, type: String })
+  lastUpdatedAt: string | null;
+}
--- a/apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.service.ts
+++ b/apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.service.ts
@@ -3,6 +3,7 @@ import { Project, Site, SiteReport, TreeSpecies } from "@terramatch-microservice
 import { DashboardQueryDto } from "./dashboard-query.dto";
 import { DashboardProjectsQueryBuilder } from "../dashboard-query.builder";
 import { Op, fn, col, literal } from "sequelize";
+import { Literal } from "sequelize/types/utils";
 
 interface DateResult {
   year: number;
@@ -19,17 +20,9 @@ export class TreeRestorationGoalService {
       }
     ]).queryFilters(query);
 
-    const projectIds: number[] = await projectsBuilder.pluckIds();
-    const projects = await Project.findAll({
-      where: { id: projectIds },
-      include: [
-        {
-          association: "organisation",
-          attributes: ["type"]
-        }
-      ]
-    });
+    const projects = await projectsBuilder.execute();
 
+    const projectIds = projects.map(project => project.id);
     const forProfitProjectIds = projects
       .filter(project => project.organisation?.type === "for-profit-organization")
       .map(project => project.id);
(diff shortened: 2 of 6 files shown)


apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.dto.ts
```diff
@@ -0,0 +1,74 @@
...
+    description: "Number of tree species for this period",
+    type: Number
+  })
+  treeSpeciesAmount: number;
+
+  @ApiProperty({
+    description: "Percentage of tree species for this period",
+    type: Number
+  })
+  treeSpeciesPercentage: number;
```
roguenet, line 23: 🎨 It's a little weird to me to provide an amount and a percentage. Not a huge deal so leave it alone if you want, but it would make more sense to me to include the amount and the goal and let the FE calculate percentage when it needs to. Then if the goal is ever desired in the future, it's already included in the response data.

apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.dto.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ApiProperty } from "@nestjs/swagger";
+import { populateDto } from "@terramatch-microservices/common/dto/json-api-attributes";
+
+class TreeRestorationData {
+  @ApiProperty({
+    description: "Due date for this restoration data",
+    type: String,
+    example: "2024-01-01T00:00:00.000000Z"
+  })
+  dueDate: string;
```
roguenet, line 11: This should be typed as a `Date`. See the createdAt / updatedAt examples in entity DTOs.

apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.dto.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  @ApiProperty({
+    description: "Total number of trees grown goal for non-profit organizations",
+    type: Number
+  })
+  nonProfitTreeCount: number;
+
+  @ApiProperty({
+    description: "Total trees grown goal across all organizations",
+    type: Number
```
roguenet, line 46: 🎨 It doesn't hurt anything, but FYI the `type` isn't needed unless the property is nullable (uses `: number | null`). In cases where the type is concrete, the swagger generation layer can correctly determine the type without it.

apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    const projectIds = projects.map(project => project.id);
+    const forProfitProjectIds = projects
+      .filter(project => project.organisation?.type === "for-profit-organization")
+      .map(project => project.id);
+
+    const nonProfitProjectIds = projects
+      .filter(project => project.organisation?.type === "non-profit-organization")
+      .map(project => project.id);
+
+    const totalTreesGrownGoal = projects.reduce((sum, project) => sum + (project.treesGrownGoal ?? 0), 0);
```
roguenet, line 34: :art: this is fine, but lodash does have a handy utility for this: `sumBy(projects, "treesGrownGoal")`

apps/dashboard-service/src/dashboard/dto/tree-restoration-goal.service.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    const approvedSitesQuery =
+      projectIds.length > 0 ? await Site.approvedIdsProjectsSubquery(projectIds) : literal("(0)");
+    const forProfitApprovedSitesQuery =
+      forProfitProjectIds.length > 0 ? await Site.approvedIdsProjectsSubquery(forProfitProjectIds) : literal("(0)");
+    const nonProfitApprovedSitesQuery =
+      nonProfitProjectIds.length > 0 ? await Site.approvedIdsProjectsSubquery(nonProfitProjectIds) : literal("(0)");
```
roguenet, line 41: These subquery methods aren't async. I'm actually a little surprised the compiler or linter isn't complaining, but these `await` should not be needed.

roguenet: I think it might be cleaner to set `undefined` instead of `literal("(0)")` for these. It will mean handling that undefined below, but at least it won't mean issuing queries to the database that are guaranteed to return nothing.

apps/dashboard-service/src/dashboard/tree-restoration-goal.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+
+@Controller("dashboard/v3/treeRestorationGoal")
+export class TreeRestorationGoalController {
+  constructor(
+    private readonly treeRestorationGoalService: TreeRestorationGoalService,
+    private readonly cacheService: CacheService
+  ) {}
+
+  @Get()
+  @JsonApiResponse([TreeRestorationGoalDto])
```
roguenet, line 18: When there's only one type of response, this doesn't need to be an array
```
@JsonApiResponse(TreeRestorationGoalDto)
```

apps/dashboard-service/src/dashboard/tree-restoration-goal.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    const timestampKey = `${cacheKey}:timestamp`;
+    const lastUpdatedAt = await this.cacheService.get(timestampKey);
+    const cachedData = await this.cacheService.get(cacheKey);
+
+    if (cachedData == null) {
+      const result = await this.treeRestorationGoalService.getTreeRestorationGoal(query);
+      const timestamp = new Date().toISOString();
+      await this.cacheService.set(cacheKey, JSON.stringify(result));
+      await this.cacheService.set(timestampKey, timestamp);
+      return result;
```
roguenet, line 31: This return is incorrect - it should be building a JSON API document.

This is optional, but a nice pattern to implement would be to add an optional factory parameter to the cache service get method. So for instance, the lines above would be something more like:
```
const lastUpdatedAt = await this.cacheService.get(timestampKey, () => new Date().toISOString());
const cachedData = await this.cacheService.get(cacheKey, async () => {
  await this.treeRestorationGoalService.getTreeRestorationGoal(query);
})
```

The get method in the cache service would then check for the data, return the cached result if it exists, and if it doesn't, await the factory method result, store that in the cache and return that result. It would streamline this usage quite a bit, and it's a fairly common cache pattern.

apps/dashboard-service/src/dashboard/tree-restoration-goal.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    const cacheKey = `dashboard:tree-restoration-goal|${this.cacheService.getCacheKeyFromQuery(query)}`;
+    const timestampKey = `${cacheKey}:timestamp`;
+    const lastUpdatedAt = await this.cacheService.get(timestampKey);
+    const cachedData = await this.cacheService.get(cacheKey);
+
+    if (cachedData == null) {
+      const result = await this.treeRestorationGoalService.getTreeRestorationGoal(query);
+      const timestamp = new Date().toISOString();
+      await this.cacheService.set(cacheKey, JSON.stringify(result));
+      await this.cacheService.set(timestampKey, timestamp);
```
roguenet, line 30: It looks like these results are set to never expire, and this controller doesn't check `lastUpdatedAt` to see if it's older than some defined timeout. Unless I'm missing something, that's going to cause this data to get calculated exactly once and then never again.

apps/dashboard-service/src/dashboard/dto/cache.service.ts
```diff
@@ -26,16 +26,27 @@ export class CacheService {
...
+          return JSON.parse(data);
+        } catch {
+          return data;
+        }
       }
+      return data;
     }
-    return data;
+
+    if (factory !== undefined && factory !== null) {
```
roguenet, line 42: Please condense to just `factory != null`

egrojMonroy: condensed now

PR #191, opened 2025-06-03
[TM-2088] Add support for trees planted entity processor

Scriptmatico:
https://gfw.atlassian.net/browse/TM-2088

### DIFF
--- a/apps/entity-service/src/entities/dto/project.dto.ts
+++ b/apps/entity-service/src/entities/dto/project.dto.ts
@@ -76,7 +76,7 @@ export class ProjectLightDto extends EntityDto {
   updatedAt: Date;
 
   @ApiProperty()
-  treesPlantedCount: number;
+  treesPlantedCount?: number;
 }
 
 export type ProjectMedia = Pick<ProjectFullDto, keyof typeof Project.MEDIA>;
--- a/apps/entity-service/src/entities/dto/project.dto.ts
+++ b/apps/entity-service/src/entities/dto/project.dto.ts
@@ -75,8 +75,8 @@ export class ProjectLightDto extends EntityDto {
   @ApiProperty()
   updatedAt: Date;
 
-  @ApiProperty()
-  treesPlantedCount?: number;
+  @ApiProperty({ nullable: true, type: Number })
+  treesPlantedCount: number | null;
 }
 
 export type ProjectMedia = Pick<ProjectFullDto, keyof typeof Project.MEDIA>;
--- a/apps/entity-service/src/entities/dto/project.dto.ts
+++ b/apps/entity-service/src/entities/dto/project.dto.ts
@@ -74,6 +74,9 @@ export class ProjectLightDto extends EntityDto {
 
   @ApiProperty()
   updatedAt: Date;
+
+  @ApiProperty()
+  treesPlantedCount: number;
 }
 
 export type ProjectMedia = Pick<ProjectFullDto, keyof typeof Project.MEDIA>;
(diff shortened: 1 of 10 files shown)


apps/entity-service/src/entities/dto/project.dto.ts
```diff
@@ -76,7 +76,7 @@ export class ProjectLightDto extends EntityDto {
   updatedAt: Date;
 
   @ApiProperty()
-  treesPlantedCount: number;
+  treesPlantedCount?: number;
```
roguenet, line 79: The convention in the DTOs is to make this `| null;` and use `{ nullable: true }` in the decorator.

apps/entity-service/src/entities/processors/entity-processor.ts
```diff
@@ -87,12 +86,31 @@ export abstract class EntityProcessor<
 
   async getLightDtos(models: ModelType[]): Promise<DtoResult<LightDto>[]> {
     const results: DtoResult<LightDto>[] = [];
+
+    let associateData: Record<number, LightDto> = {};
+
+    const allowedTypes = [Project];
+    if (allowedTypes.some(type => models[0] instanceof type)) {
+      associateData = (await this.loadAssociationData(models.map(m => String(m.id)))) as Record<number, LightDto>;
+    }
```
roguenet, line 95: Instead of this `allowedTypes` structure, let's have the default implementation just return an empty map instead of throwing an error. The base entity processor should not be responsible for knowing which subclasses have a real implementation.

apps/entity-service/src/entities/processors/entity-processor.ts
```diff
@@ -87,12 +86,31 @@ export abstract class EntityProcessor<
...
+
     for (const model of models) {
-      results.push(await this.getLightDto(model));
+      let dto = await this.getLightDto(model);
+      dto = this.mergeDto(dto, associateData, model);
```
roguenet, line 99: Instead of this merge DTO pattern, the this method should just pass the associatedData property to getLightDto, and let the subclass handle it.

apps/entity-service/src/entities/processors/project.processor.ts
```diff
@@ -291,9 +289,78 @@ export class ProjectProcessor extends EntityProcessor<
     return pTotal + sTotal + nTotal;
   }
 
-  async loadAssociationData(projectIds: number[]): Promise<number> {
-    const approvedSitesQuery = Site.approvedIdsSubquery(projectIds[0]);
-    const approvedSiteReportsQuery = SiteReport.approvedIdsSubquery(approvedSitesQuery);
-    return await TreeSpecies.visible().collection("tree-planted").siteReports(approvedSiteReportsQuery).sum("amount");
+  async loadAssociationData(projectIds: string[]): Promise<Record<number, ProjectLightDto>> {
+    const associationDtos: Record<number, ProjectLightDto> = {};
+    const numericProjectIds = projectIds.map(id => Number(id)).filter(id => !isNaN(id));
```
roguenet, line 294: This method should just accept `number[]` - no reason to strictly cast all the numbers to strings and then parse them back into numbers here.

apps/entity-service/src/entities/processors/project.processor.ts
```diff
@@ -291,9 +289,78 @@ export class ProjectProcessor extends EntityProcessor<
...
+
+    return associationDtos;
+  }
+
+  private async getTreeSpecies(approvedSiteReports: SiteReport[]) {
+    return await TreeSpecies.visible()
+      .collection("tree-planted")
+      .siteReports(approvedSiteReports.map(r => r.id))
+      .findAll({
+        attributes: ["speciesableId", [Sequelize.fn("SUM", Sequelize.col("amount")), "amount"]],
```
roguenet, line 345: Nice optimization

roguenet (approved):

PR #194, opened 2025-06-06
[TM-2084] Move demographics new controller and support querying

Scriptmatico:
https://gfw.atlassian.net/browse/TM-2084


apps/entity-service/src/entities/demographics.controller.ts
```diff
@@ -27,7 +28,12 @@ export class DemographicsController {
       await this.policyService.authorize("read", data);
       for (const demographic of data) {
         indexIds.push(demographic.uuid);
-        const demographicDto = new DemographicDto(demographic);
+        const model = LARAVEL_MODELS[demographic.demographicalType];
+        const demographicData = await model.findOne({
+          where: { id: demographic.demographicalId },
+          attributes: ["id", "uuid"]
+        });
+        const demographicDto = new DemographicDto(demographic, demographicData);
```
roguenet, line 36: This `demographicData` isn't the right shape. I'm a little surprised it compiles... it's been tricky getting the type safety guards right on those DTOs. It should look like this (example from the association processor that is also creating this DTO):
https://github.com/wri/terramatch-microservices/blob/f6ddd713bc4d42ddd8e46cc9e5cf69db7901c6cb/apps/entity-service/src/entities/processors/association-processor.ts#L72

I'm also surprised it's not complaining that `model` could be null. In this case, something like this should be right:
```
const { demographicalType: entityType, demographicalId } = demographic;
const model = LARAVEL_MODELS[entityType];
if (model == null) {
  this.logger.error('Unknown model type', entityType);
  throw new InternalServerErrorException('Unexpected demographic association type');
}
const entity = await model.findOne({ where: { id: demographicalId }, attributes: ["uuid"] });
if (entity == null) {
  this.logger.error('Demographic parent entity not found', { entityType, id: demographicalId });
  throw new NotFoundException();
}
const additionalProps = { entityType, entityUuid: entity.uuid };
const demographicDto = new DemographicDto(demographic, additionalProps);
```

PR #201, opened 2025-06-11
[TM-1554] add update bulk approval to entity update

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-1554)


roguenet (commented):
This looks kind of like you left the old implementation in place while adding the new one? I think the process bulk approval controller is not needed at all. 

Additionally, the return from the entity controller should sideload all of the affected report light DTOs so that the client gets updated data for all affected models for its cache.

roguenet:
The tasks controller already sideloads all reports on the task. Why is another controller needed that does essentially the same thing?

Scriptmatico (approved):

apps/entity-service/src/entities/dto/task-query.dto.ts
```diff
@@ -14,6 +14,14 @@ export class TaskQueryDto extends IndexQueryDto {
...
   @IsOptional()
   projectUuid?: string;
+
+  @ApiProperty({
+    required: false,
+    description:
+      "Optional filter to retrieve reports with 'nothing to report' set to true; applied only when this filter is present"
+  })
+  @IsOptional()
+  nothingToReportStatus?: boolean;
```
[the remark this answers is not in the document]
LimberHope: I had a question about the response from the update process. Should it work with sideLoad? Meaning the response should include relationships and includes, but only for the reports that were modified — is that correct?

apps/entity-service/src/entities/processors/entity-processor.ts
```diff
@@ -169,6 +176,71 @@ export abstract class EntityProcessor<
   loadAssociationData(ids: number[]): Promise<Record<number, object>> {
     return Promise.resolve({});
   }
+
+  private async findReportsByUuids<T extends ReportModel>(
+    modelClass: ModelCtor<T>,
+    uuids: string[] | undefined
+  ): Promise<T[]> {
+    if (uuids == undefined) return [];
```
Scriptmatico, line 184:
```suggestion
  private async findReportsByUuids<T extends ReportModel>(
    modelClass: ModelCtor<T>,
    uuids: string[]
  ): Promise<T[]> {
    if (uuids == null) return [];
```

I think this is better in case the value becomes null or undefined, and we force to receive a value

apps/entity-service/src/entities/entities.controller.ts
```diff
@@ -144,17 +165,24 @@ export class EntitiesController {
...
     const { id, dto } = await processor.getFullDto(model);
     document.addData(id, dto);
+
+    const filterUuidsMap: Record<string, string[]> = {};
+    if (nurseryReportUuids !== null || siteReportUuids !== null) {
+      if (siteReportUuids) filterUuidsMap["siteReports"] = siteReportUuids;
+      if (nurseryReportUuids) filterUuidsMap["nurseryReports"] = nurseryReportUuids;
+
+      await processSideloads(processor, document, model, query.sideloads, filterUuidsMap);
+    }
```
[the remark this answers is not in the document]
LimberHope: I just moved the bulkApproved method to the tasks.controller.ts. Currently, what I have in mind for proceeding with the implementation is that when performing the bulk approval, the admin user filters the tasks by project — which will return multiple tasks related to that filtered project. Then, the admin user will be able to see a list of reports from all those tasks and select which ones to approve.

The idea is to perform the bulk approval of those reports using the PATCH method on the tasks, and from the frontend, trigger the connection call for each task.

I was also thinking of keeping the relationships when retrieving the reports from the tasks to identify which task each report belongs to — mainly so that we can call the PATCH connection only on tasks that have reports with nothingToReport = true and are not already approved. Internally, the backend does validation. Right now, multiple report UUIDs are received and during the bulkApproval process, only the reports related to the UUID of the task in the request are updated.

Does this approach sound correct to you? Or were you expecting a different logic for this flow?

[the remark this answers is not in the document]
LimberHope: Thanks for the clarification. In that case, the current implementation on the BE side will continue as planned. Most of it is already in place, since we’re already set up to receive the report UUIDs from the FE

apps/entity-service/src/entities/tasks.controller.ts
```diff
@@ -41,7 +47,8 @@ export class TasksController {
 
       for (const task of tasks) {
         indexIds.push(task.uuid);
-        document.addData(task.uuid, new TaskLightDto(task));
+
+        await this.tasksService.addFullTaskDto(document, task);
```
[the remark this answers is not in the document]
LimberHope: Thanks for the feedback. Just to clarify: in this part, we’re reusing the full DTO because on the frontend we need to filter the tasks that have nothingToReport == true and are not approved. Do you think it would be better to change the approach and fetch the reports separately from the frontend?

apps/entity-service/src/entities/tasks.service.ts
```diff
@@ -154,6 +167,82 @@ export class TasksService {
...
 
+  async approveBulkReports(attributes: TaskUpdateBody, taskId: number): Promise<void> {
+    const user = await User.findOne({
+      where: { id: this.entitiesService.userId },
+      attributes: ["id", "firstName", "lastName", "emailAddress"]
+    });
+
+    const [siteReports, nurseryReports] = await Promise.all([
+      this.findReportsByUuids(SiteReport, attributes.data.attributes.siteReportNothingToReportUuid ?? [], taskId),
+      this.findReportsByUuids(NurseryReport, attributes.data.attributes.nurseryReportNothingToReportUuid ?? [], taskId)
```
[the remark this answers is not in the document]
LimberHope: The correction has been applied: approveBulkReports now accepts the full task object and uses loadReports(task) to efficiently load and update reports.

apps/entity-service/src/entities/tasks.service.ts
```diff
@@ -154,6 +167,82 @@ export class TasksService {
...
+    taskId: number
+  ): Promise<T[]> {
+    if (uuids == null) return [];
+
+    const reports = await modelClass.findAll({
+      where: { uuid: { [Op.in]: uuids }, taskId },
+      attributes: ["id", "uuid"]
+    });
+    return reports as T[];
+  }
```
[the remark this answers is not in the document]
LimberHope: removed

apps/entity-service/src/entities/tasks.service.ts
```diff
@@ -154,6 +167,82 @@ export class TasksService {
...
+      await AuditStatus.bulkCreate(auditStatusRecords);
+    }
+  }
+
+  private async findReportsByUuids<T extends ReportModel>(
+    modelClass: ModelCtor<T>,
+    uuids: string[],
+    taskId: number
+  ): Promise<T[]> {
+    if (uuids == null) return [];
```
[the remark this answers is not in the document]
LimberHope: I think I already removed that.

apps/entity-service/src/entities/tasks.controller.ts
```diff
@@ -117,6 +110,15 @@ export class TasksController {
     const task = await this.tasksService.getTask(uuid);
     await this.policyService.authorize("update", task);
 
+    if (
+      updatePayload.data.attributes.nurseryReportNothingToReportUuids != null ||
+      updatePayload.data.attributes.siteReportNothingToReportUuids != null
+    ) {
+      await this.tasksService.approveBulkReports(updatePayload.data.attributes, task);
+      const updatedTask = await this.tasksService.getTask(uuid);
+      return (await this.tasksService.addFullTaskDto(buildJsonApi(TaskFullDto), updatedTask)).serialize();
```
roguenet, line 119: Two things:
* returning the task from the service is not necessary. It's the same instance that was passed in, so it's the same value as `task`.
* Since the two returns in this method are doing the same thing, I think it would be cleaner to just break them into a simple if/else structure and use the one return:
```
const { nurseryReportNothingToReportUuids, siteReportNothingToReportUuids, status = updatePayload.data.attributes;
if (nurseryReportNothingToReportUuids != null || siteReportNothingToReportUuids != null) {
  await this.tasksService.approvedBulkReports(updatePayload.data.attributes, task);
} else if (status != null) {
  if (status === "awaiting-approval") {
    await this.tasksService.submitForApproval(task);
    await task.save();
  } else {
    throw new BadRequestException(`Status not supported by this controller: ${status}`);
  }
}

return (await this.tasksService.addFullTaskDto(buildJsonApi(TaskFullDto), task)).serialize();
```

apps/entity-service/src/entities/tasks.service.ts
```diff
@@ -167,54 +167,37 @@ export class TasksService {
...
-    taskId: number
-  ): Promise<T[]> {
-    if (uuids == null) return [];
-
-    const reports = await modelClass.findAll({
-      where: { uuid: { [Op.in]: uuids }, taskId },
-      attributes: ["id", "uuid"]
-    });
-    return reports as T[];
+    return task;
```
roguenet, line 200: See comment in the controller - returning the task here is not necessary.

apps/entity-service/src/entities/tasks.service.ts
```diff
@@ -154,6 +167,65 @@ export class TasksService {
...
+    await this.updateReportsStatus(NurseryReport, attributes.nurseryReportNothingToReportUuids ?? [], APPROVED, taskId);
+
+    await this.loadReports(task);
+
+    const siteReports = await SiteReport.findAll({
+      where: { uuid: { [Op.in]: attributes.siteReportNothingToReportUuids ?? [] }, taskId }
+    });
+    const nurseryReports = await NurseryReport.findAll({
+      where: { uuid: { [Op.in]: attributes.nurseryReportNothingToReportUuids ?? [] }, taskId }
+    });
```
roguenet, line 189: These are already loaded from the DB in `this.loadReports(task)`. So they should simply be pulled from the task, as shown in this comment: https://github.com/wri/terramatch-microservices/pull/201#discussion_r2183206887

LimberHope: the changes to filtering are actually in this part of the code [here](https://github.com/wri/wri-terramatch-website/pull/1270/files#r2183863847)

apps/entity-service/src/entities/tasks.controller.ts
LimberHope: Comment resolved.

apps/entity-service/src/entities/tasks.service.ts
LimberHope: Comment resolved.

apps/entity-service/src/entities/tasks.service.ts
```diff
@@ -167,26 +168,30 @@ export class TasksService {
...
-      where: { uuid: { [Op.in]: attributes.siteReportNothingToReportUuids ?? [] }, taskId }
-    });
-    const nurseryReports = await NurseryReport.findAll({
-      where: { uuid: { [Op.in]: attributes.nurseryReportNothingToReportUuids ?? [] }, taskId }
-    });
+    const siteReports = filter(
+      (attributes.siteReportNothingToReportUuids ?? []).map(uuid =>
+        task.siteReports?.find(siteReport => siteReport.uuid === uuid)
+      ),
+      (report): report is SiteReport => report !== undefined && report !== null
```
roguenet, line 187: This argument is not needed - the lodash `filter` method does null filtering by default. That's why I use it over `Array.filter()` for cases like this.

LimberHope: resolved

roguenet:
The coverage in the entity service unit tests has dipped below the configured minimum, as a result of this PR and several other recent PRs not having enough test coverage. A general reminder for the team: please remember to check the coverage locally (with `nx test <service> --coverage --runInBand --passWithNoTests`, and then open the report at `coverage/apps/<service>/index.html`) when writing your unit tests. Here's what this branch shows for me:
image


For this PR, please at least get your coverage on the tasks service up high enough for the test suite to pass. Would be cool if you could improve the coverage on some of the others as well, but I know you're ready to get this merged, so for now just get the coverage high enough for the testing suite to pass successfully.

roguenet (approved):

PR #206, opened 2025-06-18
[TM-2135] move boundingbox resource to research service

egrojMonroy:
https://gfw.atlassian.net/browse/TM-2135


roguenet (commented):
`api-gateway-stack.ts` needs an update for the move of the namespace (and we'll need to make sure to re-deploy the gateway)

PR #212, opened 2025-06-24
[TM-2033] Migrate jobs created endpoint v2 to v3 ms

Scriptmatico:
https://gfw.atlassian.net/browse/TM-2033


apps/dashboard-service/src/dashboard/total-jobs-created.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+  @NoBearerAuth
+  @JsonApiResponse([TotalJobsCreatedDto, DelayedJobDto])
+  @ApiOperation({ operationId: "getTotalJobsCreated", summary: "Get total jobs created" })
+  async getTotalJobsCreated(@Query() query: DashboardQueryDto) {
+    const cacheKey = `dashboard:jobs-created|${this.cacheService.getCacheKeyFromQuery(query)}`;
+    let cachedData = await this.cacheService.get(cacheKey);
+    if (cachedData == null) {
+      cachedData = await this.jobsCreatedService.getTotals(query);
+      await this.cacheService.set(cacheKey, JSON.stringify(cachedData));
+    }
```
roguenet, line 29: The cache service got an update that could be used here to improve the code a bit:
```
const data = await this.cacheService.get(cacheKey, () => this.jobsCreatedService.getTotals(query));
```
The new optional factory parameter will only be called if the cache get is a miss, and will stash the returned value before returning it.

apps/dashboard-service/src/dashboard/total-jobs-created.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+        totalMen,
+        totalNonYouth,
+        totalPt,
+        totalPtMen,
+        totalPtNonYouth,
+        totalPtWomen,
+        totalPtYouth,
+        totalWomen,
+        totalYouth
+      })
```
roguenet, line 70: Iterating all these members twice is a little obnoxious. We already know the service is returning a valid copy of the DTO, so instead of destructing `cachedData` and then creating a new object with it here, this can simply be
```
document.addData(stableQuery, new TotalJobsCreatedDto(cachedData));
```

roguenet (approved):

PR #218, opened 2025-06-30
[TM-2129] string to array cohort

cesarLima1:
https://gfw.atlassian.net/browse/TM-2129

### DIFF
--- a/apps/dashboard-service/src/dashboard/dashboard-query.builder.ts
+++ b/apps/dashboard-service/src/dashboard/dashboard-query.builder.ts
@@ -1,4 +1,4 @@
-import { Attributes, Filterable, FindOptions, Includeable, Op, OrderItem, WhereOptions } from "sequelize";
+import { Attributes, Filterable, FindOptions, Includeable, Op, OrderItem, WhereOptions, Sequelize } from "sequelize";
 import { Model, ModelCtor } from "sequelize-typescript";
 import { DashboardQueryDto } from "./dto/dashboard-query.dto";
 import { isObject, flatten, isEmpty } from "lodash";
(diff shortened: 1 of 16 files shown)


apps/dashboard-service/src/dashboard/dashboard-query.builder.ts
```diff
@@ -57,6 +55,25 @@ export class DashboardProjectsQueryBuilder<T extends Model<T> = Project> {
 
     this.where(where);
 
+    if (filters?.cohort != null && filters.cohort.length > 0) {
+      const cohortConditions = filters.cohort
+        .map(cohort => {
+          const escapedCohort = this.MODEL.sequelize?.escape(`"${cohort}"`) ?? `'"${cohort}"'`;
```
roguenet, line 61: The typing system doesn't know this, but sequelize is guaranteed to be set here. If it's not, that's a major internal error in the server. I've gotten around this kind of default setting in the past with something like this:
https://github.com/wri/terramatch-microservices/blob/deffb0bdf4926d299a41bdede4c8ebc75b51306b/libs/database/src/lib/util/subquery.builder.ts#L39-L42

You could implement a method like that in this class, and in this context you could be extra safe by doing it more like this:
```
  get sql() {
    if (this.MODEL.sequelize == null) throw new InternalServerErrorException("Model is missing sequelize conection");
    return this.MODEL.sequelize;
  }
```

In particular, having the default be to just go ahead and wrap the cohort and send it to the DB opens us up to an injection attack vector, so definitely need to avoid that.

apps/dashboard-service/src/dashboard/dashboard-query.builder.ts
```diff
@@ -57,6 +55,25 @@ export class DashboardProjectsQueryBuilder<T extends Model<T> = Project> {
...
+          return `JSON_CONTAINS(cohort, ${escapedCohort})`;
+        })
+        .join(" OR ");
+      this.where(Sequelize.literal(`(${cohortConditions})`));
+    } else {
+      const defaultCohorts = ["terrafund", "terrafund-landscapes"];
+      const defaultCohortConditions = defaultCohorts
+        .map(cohort => {
+          const escapedCohort = this.MODEL.sequelize?.escape(`"${cohort}"`) ?? `'"${cohort}"'`;
```
roguenet, line 70: Same feedback here regarding escaping the content.

apps/entity-service/src/entities/processors/project.processor.ts
```diff
@@ -33,9 +33,15 @@ export class ProjectProcessor extends EntityProcessor<
   ProjectFullDto,
   ProjectUpdateAttributes
 > {
+  readonly MODEL = Project;
```
roguenet, line 36: Is this being used?

cesarLima1: Removed. I was using it at the begining for testing and forgot to remove it. Thanks for flaging!

PR #224, opened 2025-07-03
[TM-2225] Implement automatic cache update for dashboard

cesarLima1:
https://gfw.atlassian.net/browse/TM-2225

### DIFF
--- a/apps/dashboard-service/src/dashboard/warmup/dashboard-cache-warmup.service.ts
+++ b/apps/dashboard-service/src/dashboard/warmup/dashboard-cache-warmup.service.ts
@@ -8,6 +8,20 @@ import { InjectRedis } from "@nestjs-modules/ioredis";
 import Redis from "ioredis";
 import { TMLogger } from "@terramatch-microservices/common/util/tm-logger";
 
+const generateCombinations = <T>(items: readonly T[]): T[][] => {
+  const combos: T[][] = [];
+  const count = items.length;
+  combos.push([]);
+  for (let i = 1; i < 1 << count; i++) {
+    const combo: T[] = [];
+    for (let j = 0; j < count; j++) {
+      if ((i & (1 << j)) !== 0) combo.push(items[j]);
+    }
+    combos.push(combo);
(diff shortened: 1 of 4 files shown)


apps/dashboard-service/src/dashboard/warmup/dashboard-cache-warmup.service.ts
```diff
@@ -0,0 +1,102 @@
...
+        combos.push([]);
+        for (let i = 1; i < 1 << count; i++) {
+          const combo: T[] = [];
+          for (let j = 0; j < count; j++) {
+            if ((i & (1 << j)) !== 0) combo.push(items[j]);
+          }
+          combos.push(combo);
+        }
+        return combos;
+      };
```
roguenet, line 47: There doesn't seem to be any reason to define this method within the surrounding method. Please move to the top of the file above the class definition.

apps/dashboard-service/src/dashboard/warmup/dashboard-cache-warmup.service.ts
```diff
@@ -0,0 +1,102 @@
...
+            if (processed % 10 === 0 || processed === totalCombos) {
+              this.logger.log(`Processed ${processed}/${totalCombos} combinations`);
+            }
+          }
+        }
+      }
+
+      this.logger.log(`Finished warming up ${totalCombos} combinations`);
+    } catch (error) {
+      this.logger.error(`Error during cache warmup: ${error instanceof Error ? error.message : String(error)}`);
```
roguenet, line 89: If you pass the error as a second argument, we'll also get stack traces and some additional info in Sentry

roguenet (approved):

apps/dashboard-service/src/dashboard/warmup/dashboard-cache-warmup.service.ts
cesarLima1: Done. Now the error is the second argument of the logger error.

PR #231, opened 2025-07-14
[TM-2207] add initial dashboard project code

egrojMonroy:
https://gfw.atlassian.net/browse/TM-2207

this PR includes:
Light Data Endpoint: GET /dashboard/v3/dashboardProjects
Public Access: No authentication required (using @NoBearerAuth)
Caching: Performance optimization with cache service

Pending Work
Authentication Policy: Add user access control for single project views for:
Full Data Endpoint: GET /dashboard/v3/dashboardProjects/:uuid - where we will need to send light if not registered but also transform this policy from v2:
```

 public function getIfUserIsAllowedToProject(String $uuid)
    {
        /** @var User $user */
        $user = Auth::user();
        if (is_null($user)) {
            $response = (object)[
              'allowed' => false,
            ];
        } elseif ($user->hasRole('government')) {
            $response = (object)[
              'allowed' => false,
            ];
        } elseif ($user->hasRole('funder')) {
            $verifyInUserProgram = Project::where('uuid', $uuid)
                ->where('framework_key', $user->program)
                ->exists();
            if ($verifyInUserProgram) {
                $isAllowed = $verifyInUserProgram;
            } else {
                $frameworksSlugs = $user->my_frameworks_slug;
                $verifyInUserFrameworks = Project::where('uuid', $uuid)
                    ->whereIn('framework_key', $frameworksSlugs)
                    ->exists();
                $isAllowed = $verifyInUserFrameworks;
            }
            $response = (object)[
                'allowed' => $isAllowed,
            ];
        } elseif ($user->hasRole('project-developer')) {
            $project = Project::where('uuid', $uuid)->first();
            $projectId = $project ? $project->id : null;

            $isInvite = ProjectInvite::where('email_address', $user->email_address)
                ->where('project_id', $projectId)
                ->exists();

            $isAllowedByOrganization = $project && $user->organisation
                && ($user->organisation->id == $project->organisation_id
                    || $user->projects->contains($project->id));

            $response = (object)[
                'allowed' => $isInvite || $isAllowedByOrganization,
            ];
        } elseif ($user->hasDashboardAdminAccess()) {
            $response = (object)[
                'allowed' => true,
            ];
        } else {
            $response = (object)[
                'allowed' => false,
            ];
        }

        return response()->json($response);
    }
```

### DIFF
--- a/apps/entity-service/src/entities/dto/nursery.dto.ts
+++ b/apps/entity-service/src/entities/dto/nursery.dto.ts
@@ -11,6 +11,7 @@ import {
 } from "@terramatch-microservices/database/constants/status";
 import { MediaDto } from "./media.dto";
 import { HybridSupportProps } from "@terramatch-microservices/common/dto/hybrid-support.dto";
+import { PLANTING_STATUSES, PlantingStatus } from "@terramatch-microservices/database/constants/planting-status";
 
 @JsonApiDto({ type: "nurseries" })
 export class NurseryLightDto extends EntityDto {
@@ -96,6 +97,13 @@ export class NurseryFullDto extends NurseryLightDto {
 
   @ApiProperty({ nullable: true, type: String })
   oldModel: string | null;
+
+  @ApiProperty({
+    nullable: true,
+    description: "Planting status for this nursery",
+    enum: PLANTING_STATUSES
+  })
+  plantingStatus: PlantingStatus | null;
 
   @ApiProperty({ nullable: true, type: Number })
   nurseryReportsTotal: number | null;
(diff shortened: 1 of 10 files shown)


apps/dashboard-service/src/dashboard/dto/dashboard-entity.dto.ts
```diff
@@ -0,0 +1,20 @@
+import { IsIn, IsUUID } from "class-validator";
+import { JsonApiDto } from "@terramatch-microservices/common/decorators";
+import { ApiProperty } from "@nestjs/swagger";
+import { DASHBOARD_ENTITIES, DashboardEntity } from "@terramatch-microservices/database/constants";
+
+@JsonApiDto({ type: "dashboardEntity" })
```
roguenet, line 6: Is this going to be sent to the client directly, or only as a super class? If it's a super class only, please remove the `@JsonApiDto` declaration. If it's a real entity that will be sent to the client, please make the type plural (`dashboardEntities`)

egrojMonroy: Removed as it is never sent to client because is just a parent class

PR #233, opened 2025-07-15
[TM-2207] auth project

egrojMonroy:
Add authentication for project allowed to user. 
So now, when user is allowed and request a single project, if the user is allowed the full data will be returned, otherwise will keep with light dto.


apps/dashboard-service/src/dashboard/processors/dashboard-projects.processor.ts
```diff
@@ -85,17 +91,42 @@ export class DashboardProjectsProcessor extends DashboardEntityProcessor<
       organisationType: project.organisation?.type ?? null,
       treesGrownGoal: project.treesGrownGoal,
       totalSites: totalSites,
+      is_light: true,
```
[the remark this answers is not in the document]
egrojMonroy: done

apps/dashboard-service/src/dashboard/decorators/current-user.decorator.ts
```diff
@@ -1,49 +1,8 @@
 import { createParamDecorator, ExecutionContext } from "@nestjs/common";
-import { User } from "@terramatch-microservices/database/entities";
-import { JwtService } from "@nestjs/jwt";
 
-export const CurrentUser = createParamDecorator(async (data: unknown, ctx: ExecutionContext): Promise<User | null> => {
+export const CurrentUser = createParamDecorator((data: unknown, ctx: ExecutionContext): number | null => {
```
roguenet, line 3: Let's call this `CurrentUserId` instead to make it a bit more clear.

apps/dashboard-service/src/dashboard/decorators/current-user.decorator.ts
```diff
@@ -1,49 +1,8 @@
...
-          through: { attributes: ["isManaging", "isMonitoring"] }
-        }
-      ]
-    });
-  } catch (error) {
-    return error;
-  }
+  return request.authenticatedUserId !== undefined && request.authenticatedUserId !== null
+    ? request.authenticatedUserId
+    : null;
```
roguenet, line 7: Please don't check for both undefined and null; it's overly verbose and not necessary. I also think this decorator should return `undefined` instead of `null` if there is no current user. I would shorten this whole decorator to:
```
export const CurrentUserId = createParamDecorator(_, ctx) => ctx.switchToHttp().getRequest().authenticatedUserId;
```

apps/dashboard-service/src/dashboard/dashboard-entities.controller.ts
```diff
@@ -23,7 +25,7 @@ export class DashboardEntitiesController {
   @Get(":entity")
   @NoBearerAuth
```
roguenet, line 26: I'm not familiar with interceptors - does it prevent the current auth guard from firing? If so, this `@NoBearerAuth` shouldn't be needed anymore.

apps/dashboard-service/src/dashboard/dashboard-entities.controller.ts
```diff
@@ -61,7 +63,7 @@ export class DashboardEntitiesController {
   async findOne(
     @Param("entity") entity: DashboardEntity,
     @Param("uuid") uuid: string,
-    @CurrentUser() user: User | null
+    @CurrentUser() userId: number | null
```
roguenet, line 66: With the recommendation above, this will change to `@CurrentUser() userId?: number`

apps/dashboard-service/src/dashboard/processors/dashboard-projects.processor.ts
```diff
@@ -78,22 +79,12 @@ export class DashboardProjectsProcessor extends DashboardEntityProcessor<
...
-      totalSites: totalSites,
-      is_light: true,
-      totalJobsCreated: totalJobsCreated
-    });
+    const dto = new DashboardProjectsLightDto(project, {
+      totalSites,
+      totalHectaresRestoredSum,
+      treesPlantedCount,
+      totalJobsCreated
+    } as HybridSupportProps<DashboardProjectsLightDto, Project>);
```
roguenet, line 87: Ahh, so much more concise 👍🏻

apps/dashboard-service/src/dashboard/services/dashboard-auth.service.ts
```diff
@@ -1,33 +1,26 @@
...
 @Injectable()
 export class DashboardAuthService {
   constructor(private readonly policyService: PolicyService) {}
 
-  async checkUserProjectAccess(projectUuid: string, user: User | null): Promise<DashboardAuthResult> {
-    if (user === null) {
-      return { allowed: false };
+  async userHasFullProjectAccess(projectUuid: string, userId: number | null): Promise<boolean> {
+    if (userId === null) {
+      return false;
```
roguenet, line 11: It's interesting that the user id is only needed for a null check now. The policy service will always [throw a unauthed exception if there is no current user id](https://github.com/wri/terramatch-microservices/blob/fa2f1c15ab80e1c8e3117d4be533aafab07dcab5/libs/common/src/lib/policies/policy.service.ts#L106-L107). 

Also, I think the controller should be handling the null project case (really should be throwing a 404) - this policy check should be able to assume it's dealing with a real project.

I think this whole service should go away entirely. Instead of this approach, modify the policy service to have a method that returns a boolean instead of throwing:
```
  async hasAccess(action: string, subject: Model | EntityClass | Model[]) {
    try {
      await this.authorize(action, subject);
      return true;
    } catch (e) {
      return false;
    }
  }
```

and in the controller, simply call `const authResult = this.policyService.hasAccess("read", model);`

apps/dashboard-service/src/dashboard/decorators/current-user.decorator.ts
roguenet: Update: see recommendations below: I think we can get rid of this decorator entirely.

apps/dashboard-service/src/dashboard/dashboard-entities.controller.ts
egrojMonroy: removed decorator CurrentUser, now using the policy.service.ts

apps/dashboard-service/src/dashboard/decorators/current-user.decorator.ts
egrojMonroy: Now removed

apps/dashboard-service/src/dashboard/decorators/current-user.decorator.ts
egrojMonroy: removed CurrentUser decorator

apps/dashboard-service/src/dashboard/dashboard-entities.controller.ts
egrojMonroy: No. They work together, without @NoBearerAuth it will throw "unauthorized". And after that the interceptor process the authentication if it is present. 
So the NoBearerAuth has to be present.

roguenet:
This looks ready, but there are some test failures to address.

roguenet (approved):

PR #243, opened 2025-07-22
[TM-2124] migration hectares under restoration segment to v3 fixes filters

Scriptmatico:


apps/dashboard-service/src/dashboard/hectares-restoration.service.ts
```diff
@@ -52,29 +52,34 @@ export class HectaresRestorationService {
...
             }
           ]
         }
       ],
       where: {
         status: "approved",
         isActive: true
-      }
+      },
+      paranoid: false
```
roguenet, line 82: `paranoid: false` is the default. Why was it needed in all these clauses?

apps/dashboard-service/src/dashboard/hectares-restoration.controller.ts
```diff
@@ -21,9 +20,6 @@ export class HectaresRestorationController {
   @JsonApiResponse([HectareRestorationDto])
   @ApiOperation({ operationId: "getHectaresRestoration", summary: "Get hectares restoration" })
   async getHectaresRestoration(@Query() query: DashboardQueryDto) {
-    if (isEmpty(query.projectUuid)) {
-      throw new BadRequestException("Invalid projectUuid");
-    }
```
roguenet, line 26: I believe this should also be restored - it's an error to load this endpoint without a project ID.

roguenet (approved):

PR #283, opened 2025-08-28
[TM-2381] add category filter for impact story

egrojMonroy:
https://gfw.atlassian.net/browse/TM-2381

### DIFF
--- a/apps/entity-service/src/entities/dto/impact-story-query.dto.ts
+++ b/apps/entity-service/src/entities/dto/impact-story-query.dto.ts
@@ -1,5 +1,6 @@
 import { ApiProperty } from "@nestjs/swagger";
 import { IsArray, IsOptional } from "class-validator";
+import { Type } from "class-transformer";
 import { IndexQueryDto } from "./index-query.dto";
 import { ORGANISATION_TYPES, OrganisationType } from "@terramatch-microservices/database/constants";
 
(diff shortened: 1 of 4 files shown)


apps/entity-service/src/entities/dto/impact-story-query.dto.ts
```diff
@@ -27,9 +28,11 @@ export class ImpactStoryQueryDto extends IndexQueryDto {
   @IsOptional()
   projectUuid?: string;
 
-  @ApiProperty({ required: false })
+  @ApiProperty({ required: false, isArray: true })
   @IsOptional()
-  category?: string;
+  @IsArray()
+  @Type(() => Array)
```
roguenet, line 34: Shouldn't this be `String`? Really though, I think the `@Type` annotation is not needed here. See a similar case here: https://github.com/wri/terramatch-microservices/blob/07ed297c5e8e608e9a36a032f1ac5628e2b4e6dc/apps/entity-service/src/entities/dto/entity-query.dto.ts#L46-L49

egrojMonroy: I removed the type. I think it;s ok to keep it as a string[] as it is more consistent with the category datatype in the DB.

roguenet (approved):

PR #285, opened 2025-08-29
[TM-2427] add policy for project manager in sitepolygons

egrojMonroy:


roguenet (commented):
On the surface this looks ok, but there are some spec failures to dig into.

PR #312, opened 2025-09-25
[TM-2503] feat: add patch media

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-2503)

### DIFF
--- a/apps/entity-service/src/entities/files.controller.spec.ts
+++ b/apps/entity-service/src/entities/files.controller.spec.ts
@@ -1,6 +1,6 @@
 import { serialize } from "@terramatch-microservices/common/util/testing";
 import { Test, TestingModule } from "@nestjs/testing";
-import { BadRequestException, NotFoundException, UnauthorizedException } from "@nestjs/common";
+import { NotFoundException, UnauthorizedException } from "@nestjs/common";
 import { FilesController } from "./files.controller";
 import { FileUploadService } from "../file/file-upload.service";
 import { PolicyService } from "@terramatch-microservices/common/policies/policy.service";
@@ -42,7 +42,7 @@ describe("FilesController", () => {
       getMedias: jest.fn(),
       updateMedia: jest.fn(),
       getProjectForModel: jest.fn(),
-      updateCover: jest.fn()
+      unsetMediaCoverForProject: jest.fn()
     } as unknown as jest.Mocked<MediaService>;
     mockMediaOwnerProcessor = { getBaseEntity: jest.fn() };
     entitiesService = {
@@ -147,6 +147,7 @@ describe("FilesController", () => {
       const mockedMedia = { uuid: "media-uuid" } as Media;
       mediaService.getMedia.mockResolvedValue(mockedMedia);
       policyService.authorize.mockResolvedValue();
+      mediaService.updateMedia.mockResolvedValue(mockedMedia);
       const model = { uuid: "model-uuid", id: 1 };
       (getBaseEntityByLaravelTypeAndId as jest.Mock).mockResolvedValue(model);
       await controller.mediaUpdate(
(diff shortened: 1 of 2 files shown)


libs/common/src/lib/media/media.service.ts
```diff
@@ -26,6 +36,81 @@ export class MediaService {
...
+      const nursery = await Nursery.findOne({ where: { id: model.nurseryId }, attributes: ["projectId"] });
+      project = await Project.findOne({ where: { id: nursery?.projectId } });
+    }
+    console.log("project");
+    if (project == null) throw new BadRequestException(`Media is not part of a project.`);
+    return project;
+  }
+
+  async updateCover(media: Media, project: Project) {
+    await Media.update(
```
roguenet: When we chatted about this in Slack, I believe I mentioned that the endpoint should return all updated media in the `included` array of the JSON API response so that the client side cache for these models is updated.

apps/entity-service/src/entities/files.controller.ts
```diff
@@ -107,12 +109,32 @@ export class FilesController {
...
     const model = await getBaseEntityByLaravelTypeAndId(media.modelType, media.modelId);
     await this.policyService.authorize("updateFiles", model);
+    const updatedMedia = await this.mediaService.updateMedia(media, updatePayload);
+
     if (updatePayload.data.attributes.isCover != null && updatePayload.data.attributes.isCover === true) {
       const project = await this.mediaService.getProjectForModel(model);
       await this.policyService.authorize("read", project);
-      await this.mediaService.updateCover(media, project);
+      const updatedMedias = await this.mediaService.unsetMediaCoverForProject(updatedMedia, project);
+      updatedMedia["media"] = updatedMedias.map(
```
roguenet: For this, it should not be on the main resource. JSON:API documents have a "data" member and an "included" member. The included member is the place to put sideloaded resources, which is what these are. In an earlier version of the json-api-builder, users use to have to declare if a given piece of data goes to `data` or `included`. I removed that and made it always send to `data` if the type matches, and to `included` if it doesn't. This is the first case where we want to send DTOs to `included` even though the type matches that of the document, so please start by applying this patch:
```
diff --git a/libs/common/src/lib/util/json-api-builder.ts b/libs/common/src/lib/util/json-api-builder.ts
index 34c0b2a5..c6efb033 100644
--- a/libs/common/src/lib/util/json-api-builder.ts
+++ b/libs/common/src/lib/util/json-api-builder.ts
@@ -141,10 +141,22 @@ export class DocumentBuilder {
 
   constructor(public readonly resourceType: string, public readonly options: DocumentBuilderOptions = {}) {}
 
-  addData<DTO>(id: string, attributes: DTO): ResourceBuilder {
+  /**
+   * Adds data to the final JSON:API document. If the type of the resource does not match the declared
+   * type of the document builder, it will be included in the `included` array of the document.
+   *
+   * If the goal is to send the data to the `included` array even if it does match, use the forceIncluded
+   * parameter
+   *
+   * @param id String ID of the resource (usually the UUID)
+   * @param attributes The DTO attributes for the resource
+   * @param forceIncluded Set to true if you wish this data to go to the included array even if it
+   *   matches the declared type of the document builder. Defaults to false.
+   */
+  addData<DTO>(id: string, attributes: DTO, forceIncluded = false): ResourceBuilder {
     const builder = new ResourceBuilder(id, attributes as Attributes, this);
 
-    if (builder.type === this.resourceType) {
+    if (forceIncluded || builder.type === this.resourceType) {
       const collision = this.data.find(({ id: existingId }) => existingId === id);
       if (collision != null) {
         throw new ApiBuilderException(`This resource is already in data [${id}]`);
```
And then adding the DTOs for these updated media via 
```
for (const media of updatedMedias) {
  document.addData(media.uuid, new MediaDto(...), true);
}
```

roguenet (approved):

LimberHope (approved):

PR #314, opened 2025-09-26
[TM-2478] Polygon Size and Estimated Area validators

cesarLima1:
https://gfw.atlassian.net/browse/TM-2478

### DIFF
--- a/apps/research-service/src/validations/validation.service.spec.ts
+++ b/apps/research-service/src/validations/validation.service.spec.ts
@@ -9,6 +9,7 @@ import {
   Site,
   Project
 } from "@terramatch-microservices/database/entities";
+import { Literal } from "sequelize/types/utils";
 import { ValidationType } from "@terramatch-microservices/database/constants";
 
 interface MockCriteriaSite {
@@ -50,7 +51,8 @@ jest.mock("@terramatch-microservices/database/entities", () => ({
     sum: jest.fn()
   },
   Site: {
-    findAll: jest.fn()
+    findAll: jest.fn(),
+    uuidsSubquery: jest.fn()
   },
   Project: {
     findByPk: jest.fn()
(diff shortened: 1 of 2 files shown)


apps/research-service/src/validations/validators/estimated-area.validator.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  private async generateAreaDataProject(sitePolygon: SitePolygon): Promise<{
+    valid: boolean;
+    sumAreaProject: number | null;
+    percentageProject: number | null;
+    totalAreaProject: number | null;
+    lowerBoundProject: number | null;
+    upperBoundProject: number | null;
+  }> {
+    const site = sitePolygon.site != null ? sitePolygon.site : await sitePolygon.loadSite();
```
roguenet, line 132: The `loadSite` method does these checks for you, so this can simply be
```
const site = await sitePolygon.loadSite();
```
If it's already been loaded, it'll return immediately.

apps/research-service/src/validations/validators/estimated-area.validator.ts
```diff
@@ -0,0 +1,198 @@
...
+    });
+
+    if (sites.length === 0) {
+      return 0;
+    }
+
+    const siteUuids = sites.map(site => site.uuid);
+    const result = await SitePolygon.sum("calcArea", {
+      where: {
+        siteUuid: siteUuids,
```
roguenet, line 192: This should be handled with a subquery instead of the `findAll` above. Should this be using only approved sites? If so, there's one already ready for you to use (`approvedUuidsSubquery`). If not, you can add this to `Site` and use it:
```
  static uuidsSubquery(projectId: number) {
    return Subquery.select(Site, "uuid").eq("projectId", projectId).literal;
  }
```

roguenet (approved):

PR #315, opened 2025-09-26
[TM-2480] add update indicators for financial report

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-2480)

### DIFF
--- a/apps/entity-service/src/entities/processors/financial-report.processor.ts
+++ b/apps/entity-service/src/entities/processors/financial-report.processor.ts
@@ -1,4 +1,10 @@
-import { FinancialReport, FundingType, FinancialIndicator, Media } from "@terramatch-microservices/database/entities";
+import {
+  FinancialReport,
+  FundingType,
+  FinancialIndicator,
+  Media,
+  Organisation
+} from "@terramatch-microservices/database/entities";
 import { ReportProcessor } from "./entity-processor";
 import { EntityQueryDto } from "../dto/entity-query.dto";
 import { BadRequestException } from "@nestjs/common";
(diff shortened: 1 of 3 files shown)


apps/entity-service/src/entities/processors/financial-report.processor.ts
```diff
@@ -23,6 +29,57 @@ export class FinancialReportProcessor extends ReportProcessor<
   readonly LIGHT_DTO = FinancialReportLightDto;
   readonly FULL_DTO = FinancialReportFullDto;
 
+  /**
+   * Specific method for FinancialReport custom logic. This is called automatically when nothingToReport is updated
+   */
+  protected async processFinancialReportSpecificLogic(model: FinancialReport): Promise<void> {
+    const organisation = await Organisation.findByPk(model.organisationId);
+    if (organisation == null) {
+      console.warn(`Organisation not found for FinancialReport ${model.uuid}`);
```
roguenet, line 38: Please don't use `console`. Look for other classes that create an instance of TMLogger for an example.

apps/entity-service/src/entities/processors/financial-report.processor.ts
```diff
@@ -23,6 +29,57 @@ export class FinancialReportProcessor extends ReportProcessor<
...
+    const reportIndicators = await FinancialIndicator.financialReport(model.id).findAll();
+    const existingOrgIndicators = await FinancialIndicator.organisation(organisation.id).findAll();
+
+    const orgIndicatorMap = new Map<string, FinancialIndicator>();
+    existingOrgIndicators.forEach(indicator => {
+      const key = `${indicator.year}-${indicator.collection}`;
+      orgIndicatorMap.set(key, indicator);
+    });
+
+    for (const reportIndicator of reportIndicators) {
```
roguenet, line 59: Is this likely to be a lot of indicators? Any time I see `create` in a loop, I want to move it to a [bulk create](https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#creating-in-bulk). Since this look covers both create and update then you'd need to split into bulk create and [bulk update](https://sequelize.org/docs/v7/querying/update/#updating-in-bulk). I would only bother if this is likely to be more than 10 or in any given request.

roguenet: Oh shoot, I didn't realize how the bulk update works. I was thinking it was more like an upsert where you can update multiple different rows with different values at once, but that doesn't seem to be the case. I'll go ahead and approve, but in the future, I would have left your code similar to how it was before for the update path. The only change would be to push the update promises to an array instead of `await`ing them inline, and then just await the Promise.all after the loop is complete.

roguenet (approved):

PR #632, opened 2026-03-22
[TM-3128] Migration Remove audit logs from linked reports to the report being used

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3128)

### DIFF
--- a/apps/entity-service/src/repl/oneOff/2026-03-17-cleanupReportAuditLogs.ts
+++ b/apps/entity-service/src/repl/oneOff/2026-03-17-cleanupReportAuditLogs.ts
@@ -1,0 +1,255 @@
+import { withoutSqlLogs } from "@terramatch-microservices/common/util/repl/without-sql-logs";
+import {
+  Audit,
+  Nursery,
+  NurseryReport,
+  Project,
+  ProjectReport,
+  Site,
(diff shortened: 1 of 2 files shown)


apps/entity-service/src/repl/oneOff/2026-03-17-cleanupReportAuditLogs.ts
```diff
@@ -0,0 +1,255 @@
...
+        auditableId: { [Op.in]: ids.length > 0 ? ids : [-1] }
+      }))
+    });
+  }
+
+  let cursor = 0;
+  let processed = 0;
+
+  let hasMore = true;
+  while (hasMore) {
```
roguenet, line 186: This is already written so I won't recommend you change it, but for future reference, we have an easier way to batch process models in one offs using `PaginatedQueryBuilder` and the `batchFindAll` utility special made for REPL code: https://github.com/wri/terramatch-microservices/blob/ee6ffa9b22b222b837bfe228640d8156c11cff90/apps/entity-service/src/repl/oneOff/2026-02-25-migrateProjectCountry.ts#L10-L17

roguenet (approved):

PR #633, opened 2026-03-22
[TM-3121] Add email processing for task digest and weekly polygon upd…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3121)

### DIFF
--- a/apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
+++ b/apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
@@ -3,7 +3,7 @@ import { Cron, CronExpression } from "@nestjs/schedule";
 import { ScheduledJob, Task } from "@terramatch-microservices/database/entities";
 import { ENTERPRISES, LANDSCAPES } from "@terramatch-microservices/database/constants";
 import { AWAITING_APPROVAL, APPROVED } from "@terramatch-microservices/database/constants/status";
-import { Op, Transaction } from "sequelize";
+import { Op, QueryTypes, Transaction } from "sequelize";
 import {
   REPORT_REMINDER,
   SITE_AND_NURSERY_REMINDER,
@@ -15,13 +15,17 @@ import { Queue } from "bullmq";
 import { InjectQueue } from "@nestjs/bullmq";
 import { DateTime } from "luxon";
 import { REPORT_REMINDER_EVENT, SITE_AND_NURSERY_REMINDER_EVENT, TASK_DUE_EVENT } from "./scheduled-jobs.processor";
-import {
-  TaskDigestEmail,
-  WeeklyPolygonUpdateEmail
-} from "@terramatch-microservices/common/email/terrafund-report-reminder.email";
+import { TaskDigestEmail } from "@terramatch-microservices/common/email/terrafund-report-reminder.email";
+import { WeeklyPolygonUpdateEmail } from "@terramatch-microservices/common/email/weekly-polygon-update.email";
+import { batchFindAll } from "@terramatch-microservices/common/util/batch-find-all";
+import { PaginatedQueryBuilder } from "@terramatch-microservices/common/util/paginated-query.builder";
 
 const TASK_DIGEST_CHUNK_SIZE = 100;
 const POLYGON_DIGEST_CHUNK_SIZE = 50;
+
+/** MySQL/MariaDB named locks so only one job-service replica runs each weekly cron at a time. */
+const MYSQL_LOCK_TASK_DIGEST_WEEKLY = "tm_job_svc_task_digest_wk";
+const MYSQL_LOCK_POLYGON_UPDATES_WEEKLY = "tm_job_svc_polygon_update_wk";
 
 @Injectable()
 export class ScheduledJobsService {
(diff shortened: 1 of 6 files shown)


apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -103,4 +114,50 @@ export class ScheduledJobsService {
       }
     }
   }
+
+  /**
+   * Laravel send-daily-digest-notifications: Monday 17:00 Europe/Sofia.
+   * Multiple job-service replicas may each fire this cron; use a single replica or a distributed
+   * lock if duplicate enqueues must be avoided (Laravel onOneServer).
+   */
+  @Cron("0 17 * * 1", { name: "taskDigestWeekly", timeZone: "Europe/Sofia" })
```
roguenet: I don't think there's a good reason to specify the time zone here. And if you need to, this is an odd choice - better would be UTC (or "Europe/London")

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -103,4 +114,50 @@ export class ScheduledJobsService {
...
+    while (hasMore) {
+      const tasks = await Task.findAll({
+        where: {
+          id: { [Op.gt]: cursor },
+          status: { [Op.notIn]: [AWAITING_APPROVAL, APPROVED] }
+        },
+        attributes: ["id"],
+        order: [["id", "ASC"]],
+        limit: TASK_DIGEST_CHUNK_SIZE
+      });
```
roguenet, line 142: Please use the paginated query builder instead of managing batches manually. the `batchFindAll` utility was built for REPLs, but it should work fine for this as well.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -103,4 +114,50 @@ export class ScheduledJobsService {
...
+      cursor = tasks[tasks.length - 1].id;
+      const taskIds = tasks.map(({ id }) => id);
+      total += taskIds.length;
+      await this.emailQueue.add(TaskDigestEmail.NAME, { taskIds });
+    }
+    this.logger.log(`Task digest: enqueued chunks covering up to ${total} incomplete tasks`);
+  }
+
+  /** Laravel send-weekly-polygon-update-notifications: Monday 00:00 Europe/Sofia. */
+  @Cron("0 0 * * 1", { name: "weeklyPolygonUpdates", timeZone: "Europe/Sofia" })
```
roguenet: Same here, I'd just leave the time zone off.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -103,4 +114,50 @@ export class ScheduledJobsService {
       }
     }
   }
+
+  /**
+   * Laravel send-daily-digest-notifications: Monday 17:00 Europe/Sofia.
+   * Multiple job-service replicas may each fire this cron; use a single replica or a distributed
+   * lock if duplicate enqueues must be avoided (Laravel onOneServer).
```
roguenet: I think this is definitely true for these tasks - please look into making sure it only runs on one node. Let me know if you need help tracking down how to do this.

libs/common/src/lib/email/terrafund-report-reminder.email.ts
```diff
@@ -59,3 +75,319 @@ export class TerrafundReportReminderEmail extends ProjectEmailSender {
     );
   }
 }
+
+// --- Task digest (Laravel SendDailyDigestNotificationsJob / TaskDigestMail) ---
```
roguenet: I don't think these laravel comments sprinkled through this PR are especially useful. A month from now we won't need to care about laravel anymore and a year from now this comment will just be confusing.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -103,4 +114,50 @@ export class ScheduledJobsService {
...
+
+  /** Laravel send-weekly-polygon-update-notifications: Monday 00:00 Europe/Sofia. */
+  @Cron("0 0 * * 1", { name: "weeklyPolygonUpdates", timeZone: "Europe/Sofia" })
+  async enqueueWeeklyPolygonUpdateEmails(): Promise<void> {
+    this.logger.log("Enqueueing weekly polygon update email jobs");
+    const weekAgo = DateTime.now().minus({ days: 7 }).toJSDate();
+    const uuids = await WeeklyPolygonUpdateEmail.loadRecentSitePolygonUuids(weekAgo);
+    for (let i = 0; i < uuids.length; i += POLYGON_DIGEST_CHUNK_SIZE) {
+      const sitePolygonUuids = uuids.slice(i, i + POLYGON_DIGEST_CHUNK_SIZE);
+      await this.emailQueue.add(WeeklyPolygonUpdateEmail.NAME, { sitePolygonUuids });
```
roguenet: Instead of adding to the email queue directly, please update these to use the `sendLater` mechanic instituted for other emails. Method provided by `EmailSender`, and used like so: https://github.com/wri/terramatch-microservices/blob/3a3418aca6e360115bd430ba4e0e411d633a93a2/apps/user-service/src/organisations/organisations.controller.ts#L156-L159

That mechanic gives better type safety for email parameters because the constructor for the email ensures it.

libs/common/src/lib/email/terrafund-report-reminder.email.ts
```diff
@@ -59,3 +75,319 @@ export class TerrafundReportReminderEmail extends ProjectEmailSender {
...
+  body: "terrafund-polygon-update.body",
+  cta: "terrafund-polygon-update.cta"
+};
+
+const escapeHtml = (value: string | null | undefined): string => {
+  if (value == null || value.length === 0) return "";
+  return value.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;").replace(/"/g, "&quot;");
+};
+
+export class WeeklyPolygonUpdateEmail extends EmailSender<WeeklyPolygonUpdateEmailData> {
```
roguenet: This should be in its own file.

libs/database/src/lib/entities/polygon-updates.entity.ts
```diff
@@ -64,4 +66,12 @@ export class PolygonUpdates extends Model<PolygonUpdates> {
...
   @Column(STRING)
   newStatus: string | null;
+
+  @CreatedAt
+  @Column(DATE)
+  override createdAt: Date;
+
+  @UpdatedAt
+  @Column(DATE)
+  override updatedAt: Date;
```
roguenet: Why were these added? They are provided by Sequelize and should not be specified manually. If you just needed a non-nullable copy, you can use `update.createdAt as Date` in the code that accesses it.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
Marcolr-dev: Thanks agreed. We removed timeZone from the cron so scheduling uses the process default (typically UTC on servers).

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
Marcolr-dev: Done. Task digest enqueue now uses PaginatedQueryBuilder + batchFindAll instead of a manual cursor/findAll loop.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
Marcolr-dev: Same change: timeZone removed from this cron as well.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
Marcolr-dev: Implemented. Both weekly jobs run inside a DB transaction and acquire a MySQL named lock (GET_LOCK / RELEASE_LOCK) on the same connection so only one replica enqueues at a time when multiple job-service instances are up.

libs/common/src/lib/email/terrafund-report-reminder.email.ts
Marcolr-dev: Removed those comments; the file header now only describes the v3 behaviour.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
Marcolr-dev: Updated to use new TaskDigestEmail(...).sendLater(this.emailQueue) and new WeeklyPolygonUpdateEmail(...).sendLater(this.emailQueue) for typed payloads

libs/common/src/lib/email/terrafund-report-reminder.email.ts
Marcolr-dev: Done — moved to libs/common/src/lib/email/weekly-polygon-update.email.ts and trimmed imports in terrafund-report-reminder.email.ts

libs/database/src/lib/entities/polygon-updates.entity.ts
Marcolr-dev: Removed the explicit timestamp columns; we rely on Sequelize’s default timestamp handling for this model.

roguenet (approved):

PR #634, opened 2026-03-24
[TM-2876] add initial csv export

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-2876)

**Notes for Devs**
I found when I was doing this that the v3 docker setup doesn't work for our local S3 uploads in MinIO automatically. 
This affects both export testing and any kind of file upload from the FE. To get it working, first run `docker compose down minio` and `docker compose up minio`, then follow these new directions from the README:
- To set up MinIO (local S3 proxy) using creds from the example env (`.env.local.sample`):

  - This sets up correct local bucket in the correct region for the node services to access
    ```
    $ docker compose exec minio bash
    ... once in the container
    $ mc alias set local http://localhost:9000 AKIABUVWH1HUD7YQZQAR PVMlDMep3/jLSz9GxPV3mTvH4JZynkf2BFeTu+i8
    $ mc mb local/wri
    ```


roguenet:
@pachonjcl please do the review on this one. You opened the original PR though, so you won't be able to approve. Once you're happy with it, comment here and I'll add an approval for merge.

PR #635, opened 2026-03-24
[TM-3056] ANR plots

cesarLima1:
https://gfw.atlassian.net/browse/TM-3056


apps/research-service/src/site-polygons/dto/anr-plot-geometry.dto.ts
```diff
@@ -0,0 +1,61 @@
...
+import { AnrPlotGeometry } from "@terramatch-microservices/database/entities";
+import { FeatureCollection, Geometry } from "geojson";
+
+@JsonApiDto({ type: "anrPlotGeometries" })
+export class AnrPlotGeometryDto {
+  constructor(record: AnrPlotGeometry) {
+    this.sitePolygonUuid = record.sitePolygonUuid;
+    this.geojson = record.geojson as FeatureCollection;
+    this.plotCount = record.plotCount;
+    this.createdBy = record.createdBy;
```
roguenet: Any reason this can't use `populateDto` like most of our DTOs?

apps/research-service/src/site-polygons/dto/anr-plot-geometry.dto.ts
```diff
@@ -0,0 +1,61 @@
...
+                area_m2: { type: "integer", example: 897 },
+                select: { type: "string", nullable: true, example: "Yes" }
+              }
+            }
+          }
+        }
+      }
+    }
+  })
+  geojson: FeatureCollection<Geometry, { plot_id?: number; area_m2?: number; select?: string | null } | null>;
```
roguenet: Do these need to be snake case? We're moving toward camel case in our geometry properties as well I believe, so if this is net new it would be nice to start there.

apps/research-service/src/site-polygons/anr-plot-geometry.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const plot = await this.anrPlotGeometryService.getPlotOrThrow(sitePolygonUuid);
+
+    await this.policyService.authorize("read", plot);
+
+    const document = buildJsonApi(AnrPlotGeometryDto);
+    return document.addData(sitePolygonUuid, new AnrPlotGeometryDto(plot));
+  }
+
+  @Get("geojson")
```
roguenet, line 60: I'm not sure I see the need for this endpoint? Maybe there's a strong reason for it in the FE (haven't read that yet), but since the base endpoint returns the geojson as part of its payload, it seems like we should just that for everything...

apps/research-service/src/site-polygons/anr-plot-geometry.service.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    this.assertSitePolygonEligibleForAnrPlotGeometry(sitePolygon);
+    return sitePolygon;
+  }
+
+  async getPlot(sitePolygonUuid: string): Promise<AnrPlotGeometry | null> {
+    return AnrPlotGeometry.findOne({ where: { sitePolygonUuid } });
+  }
+
+  async getPlotOrThrow(sitePolygonUuid: string): Promise<AnrPlotGeometry> {
```
roguenet: :art: I like "require" as a name for methods like this (`requirePlot`)

apps/research-service/src/site-polygons/anr-plot-geometry.service.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  async upsertPlot(
+    sitePolygonUuid: string,
+    featureCollection: FeatureCollection,
+    userId: number
+  ): Promise<AnrPlotGeometry> {
+    if (AnrPlotGeometry.sequelize == null) {
+      throw new InternalServerErrorException("Database connection not available");
+    }
```
roguenet: Please use the `sql` getter pattern found on other models: https://github.com/wri/terramatch-microservices/blob/c2ea582d258c2b7977a31097b955e28f940e9733/libs/database/src/lib/entities/polygon-geometry.entity.ts#L9-L14. This check then goes away and the lines below use `AnyPlotGeometry.sql` instead of `.sequelize`

libs/database/src/lib/entities/anr-plot-geometry.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+  InferAttributes,
+  InferCreationAttributes,
+  JSON,
+  UUID,
+  UUIDV4
+} from "sequelize";
+import { SitePolygon } from "./site-polygon.entity";
+
+@Table({ tableName: "anr_plot_geometry", underscored: true, paranoid: true })
+export class AnrPlotGeometry extends Model<InferAttributes<AnrPlotGeometry>, InferCreationAttributes<AnrPlotGeometry>> {
```
roguenet, line 26: I mentioned this on the PHP PR for the migration, but two updates here:
* Please make the table name plural (`anr_plot_geometries`) 
* Please use the ID column instead of UUID for foreign keys; we're going to be doing some work in the future to make this the norm across all relationships.

apps/research-service/src/site-polygons/anr-plot-geometry.controller.ts
egrojMonroy: Even though the regular GET already includes the GeoJSON.
We added that endpoint because there is a shared download pipeline, that uses GeoJsonExportDto. 
Should we use the regular get for the download?

apps/research-service/src/site-polygons/dto/anr-plot-geometry.dto.ts
egrojMonroy: updated

apps/research-service/src/site-polygons/anr-plot-geometry.service.ts
egrojMonroy: changed

apps/research-service/src/site-polygons/anr-plot-geometry.service.ts
egrojMonroy: we are using sql now

libs/database/src/lib/entities/anr-plot-geometry.entity.ts
egrojMonroy: updated API and here

apps/research-service/src/site-polygons/anr-plot-geometry.controller.ts
roguenet: Nah, I suppose this is OK.

roguenet (commented):
One question remaining re snake case properties in the geojson

roguenet (approved):

PR #637, opened 2026-03-26
[TM-3133] Update project and site configurations to modify land tenur…

Marcolr-dev:
[Task Linlk](https://gfw.atlassian.net/browse/TM-3133)

### DIFF
--- a/libs/common/src/lib/linkedFields/project.configuration.ts
+++ b/libs/common/src/lib/linkedFields/project.configuration.ts
@@ -32,7 +32,7 @@ export const ProjectConfiguration: LinkedFieldConfiguration<Project> = {
       label: "Land tenure",
       inputType: "select-image",
       multiChoice: true,
-      optionListKey: "land-tenures"
+      optionListKey: "land-tenures-brazil"
     },
     "pro-country": {
       property: "country",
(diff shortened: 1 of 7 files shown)


libs/common/src/lib/linkedFields/project.configuration.ts
```diff
@@ -32,7 +32,7 @@ export const ProjectConfiguration: LinkedFieldConfiguration<Project> = {
       label: "Land tenure",
       inputType: "select-image",
       multiChoice: true,
-      optionListKey: "land-tenures"
+      optionListKey: "land-tenures-brazil"
```
roguenet: I'm not sure I understand this change. What is it accomplishing? Shouldn't the base `land-tenures` list be getting changed instead of creating a new one? It's odd to me to have a country specific option list in the linked fields config because these are available to all frameworks.

Marcolr-dev: Good catch agreed, I reverted this change so pro-land-tenure-proj-area remains on the shared land-tenures option list.
For this ticket, we're only moving land_tenure_approach from project to site and updating site-linked field wiring; we’re not introducing a framework/country-specific option list at the linked-fields config level.

roguenet (approved):

PR #640, opened 2026-03-26
Patch geoparquet backup polyid

kovaca:
Changes proposed to address cross-account geoparquet write and to use a different polygon id concept (uuid), more details at https://app.asana.com/1/25496124013636/task/1213635222085401/comment/1213746614966116?focus=true

See conversation if possible, the destination bucket policy is referenced in the Asana ticket


roguenet:
Mentioned in slack, but current status is that the IAM role on our geoserver does not have access to the bucket included in the script update. In the meantime, I can apply the geoparquet SQL change and it would be reflected in the current bucket where these go in our account. Just lmk.

roguenet (approved):
Appears to be fully working now, and I've sent these changes to the bastion host for daily run.

PR #642, opened 2026-03-27
[TM-3115] remove transactional prop and unsubscribe in emails

egrojMonroy:
https://gfw.atlassian.net/browse/TM-3115

### DIFF
--- a/libs/common/src/lib/email/organisation-join-request.email.ts
+++ b/libs/common/src/lib/email/organisation-join-request.email.ts
@@ -57,11 +57,7 @@ export class OrganisationJoinRequestEmail extends EmailSender<OrganisationJoinRe
               title: "organisation-user-join-requested.title",
               body: "organisation-user-join-requested.body"
             },
-            {
-              additionalValues: {
-                transactional: "transactional"
-              }
-            }
+            {}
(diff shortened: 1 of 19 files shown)


libs/common/src/lib/email/organisation-join-request.email.ts
```diff
@@ -57,11 +57,7 @@ export class OrganisationJoinRequestEmail extends EmailSender<OrganisationJoinRe
               title: "organisation-user-join-requested.title",
               body: "organisation-user-join-requested.body"
             },
-            {
-              additionalValues: {
-                transactional: "transactional"
-              }
-            }
+            {}
```
roguenet: This fourth argument to `sendI18nTemplateEmail` is optional, so in these cases where you're now passing in an empty object, please just leave the argument off.

roguenet (commented):
One small tweak to make.

libs/common/src/lib/email/organisation-join-request.email.ts
egrojMonroy: Changed in remaining places

roguenet (approved):

PR #664, opened 2026-04-06
[TM-3119] Add cron jobs for removing stale verifications, password re…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3119)

### DIFF
--- a/apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
+++ b/apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
@@ -1,4 +1,5 @@
-import { Injectable } from "@nestjs/common";
+import { Injectable, InternalServerErrorException } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
 import { Cron, CronExpression } from "@nestjs/schedule";
 import {
   Notification,
@@ -17,6 +18,7 @@ import {
 } from "@terramatch-microservices/database/constants/scheduled-jobs";
 import type { TaskDue } from "@terramatch-microservices/database/constants/scheduled-jobs";
 import { TMLogger } from "@terramatch-microservices/common/util/tm-logger";
+import { FileService } from "@terramatch-microservices/common/file/file.service";
 import { Queue } from "bullmq";
 import { InjectQueue } from "@nestjs/bullmq";
 import { DateTime } from "luxon";
@@ -25,28 +27,20 @@ import { TaskDigestEmail } from "@terramatch-microservices/common/email/terrafun
 import { WeeklyPolygonUpdateEmail } from "@terramatch-microservices/common/email/weekly-polygon-update.email";
 import { batchFindAll } from "@terramatch-microservices/common/util/batch-find-all";
 import { PaginatedQueryBuilder } from "@terramatch-microservices/common/util/paginated-query.builder";
-import type { Dirent } from "fs";
-import * as fs from "fs/promises";
-import * as path from "path";
 
 const TASK_DIGEST_CHUNK_SIZE = 100;
 const POLYGON_DIGEST_CHUNK_SIZE = 50;
 
-/** MySQL/MariaDB named locks so only one job-service replica runs each cron at a time. */
+/** MySQL/MariaDB named locks so only one job-service replica runs each digest cron at a time. */
 const MYSQL_LOCK_TASK_DIGEST_WEEKLY = "tm_job_svc_task_digest_wk";
 const MYSQL_LOCK_POLYGON_UPDATES_WEEKLY = "tm_job_svc_polygon_update_wk";
-const MYSQL_LOCK_RM_VERIFICATIONS = "tm_job_svc_rm_verifications";
-const MYSQL_LOCK_RM_PASSWORD_RESETS = "tm_job_svc_rm_password_resets";
-const MYSQL_LOCK_RM_EXPORT_FILES = "tm_job_svc_rm_export_files";
-const MYSQL_LOCK_RM_NOTIFICATIONS = "tm_job_svc_rm_notifications";
 
 const VERIFICATION_RETENTION_HOURS = 48;
 const PASSWORD_RESET_RETENTION_DAYS = 7;
 const EXPORT_FILE_RETENTION_DAYS = 1;
+/** S3 prefix for transient export files (matches legacy public/temp layout). */
+const EXPORT_TEMP_S3_PREFIX = "temp/";
 const NOTIFICATION_RETENTION_DAYS = 90;
-const PUBLIC_STORAGE_ROOT_ENV = "PUBLIC_STORAGE_ROOT";
-
-let hasWarnedMissingPublicStorageRoot = false;
 
 @Injectable()
 export class ScheduledJobsService {
@@ -54,13 +48,14 @@ export class ScheduledJobsService {
 
   constructor(
     @InjectQueue("scheduled-jobs") private readonly scheduledJobsQueue: Queue,
-    @InjectQueue("email") private readonly emailQueue: Queue
+    @InjectQueue("email") private readonly emailQueue: Queue,
+    private readonly configService: ConfigService,
+    private readonly fileService: FileService
   ) {}
 
   @Cron(CronExpression.EVERY_5_MINUTES)
   async processScheduledJobs() {
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const transaction = await ScheduledJob.sequelize!.transaction();
+    const transaction = await ScheduledJob.sql.transaction();
     try {
       const jobs = await ScheduledJob.findAll({
         where: { executionTime: { [Op.lte]: new Date() } },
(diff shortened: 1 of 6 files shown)


apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -19,13 +25,28 @@ import { TaskDigestEmail } from "@terramatch-microservices/common/email/terrafun
...
 const POLYGON_DIGEST_CHUNK_SIZE = 50;
 
-/** MySQL/MariaDB named locks so only one job-service replica runs each weekly cron at a time. */
+/** MySQL/MariaDB named locks so only one job-service replica runs each cron at a time. */
 const MYSQL_LOCK_TASK_DIGEST_WEEKLY = "tm_job_svc_task_digest_wk";
 const MYSQL_LOCK_POLYGON_UPDATES_WEEKLY = "tm_job_svc_polygon_update_wk";
+const MYSQL_LOCK_RM_VERIFICATIONS = "tm_job_svc_rm_verifications";
+const MYSQL_LOCK_RM_PASSWORD_RESETS = "tm_job_svc_rm_password_resets";
+const MYSQL_LOCK_RM_EXPORT_FILES = "tm_job_svc_rm_export_files";
+const MYSQL_LOCK_RM_NOTIFICATIONS = "tm_job_svc_rm_notifications";
```
roguenet: I think I mentioned this in planning; these don't need locks - there's no downside risk to them accidentally running too often.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
     });
   }
 
+  @Cron(CronExpression.EVERY_5_MINUTES, { name: "removeStaleVerifications" })
+  async removeStaleVerifications(): Promise<void> {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    await Task.sequelize!.transaction(async transaction => {
```
roguenet: Please don't avoid this eslint rule (in specs is OK, but not in service code). Instead, use the `get sql()` pattern established in [other entities](https://github.com/wri/terramatch-microservices/blob/staging/libs/database/src/lib/entities/polygon-geometry.entity.ts#L9-L14). It's cleaner in the code and generates a much better error for the logs / sentry if something goes wrong.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
...
+      await this.runWithMysqlNamedLock(MYSQL_LOCK_RM_VERIFICATIONS, transaction, async () => {
+        const cutoff = DateTime.utc().minus({ hours: VERIFICATION_RETENTION_HOURS }).toJSDate();
+        const removed = await Verification.destroy({
+          where: { createdAt: { [Op.lte]: cutoff } },
+          transaction
```
roguenet: I'm not sure I see any benefit to putting this single query in a transaction.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
...
+
+  @Cron(CronExpression.EVERY_5_MINUTES, { name: "removeStalePasswordResets" })
+  async removeStalePasswordResets(): Promise<void> {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    await Task.sequelize!.transaction(async transaction => {
+      await this.runWithMysqlNamedLock(MYSQL_LOCK_RM_PASSWORD_RESETS, transaction, async () => {
+        const cutoff = DateTime.utc().minus({ days: PASSWORD_RESET_RETENTION_DAYS }).toJSDate();
+        const removed = await PasswordReset.destroy({
+          where: { createdAt: { [Op.lte]: cutoff } },
+          transaction
```
roguenet: Same here; no transaction please.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
...
+        if (removed > 0) {
+          this.logger.log(`Removed ${removed} stale password resets (older than ${PASSWORD_RESET_RETENTION_DAYS}d)`);
+        }
+      });
+    });
+  }
+
+  @Cron(CronExpression.EVERY_DAY_AT_MIDNIGHT, { name: "removeOldExportFiles" })
+  async removeOldExportFiles(): Promise<void> {
+    const publicRoot = process.env[PUBLIC_STORAGE_ROOT_ENV];
```
roguenet: This needs to be retrieved from the config service. `process.env` is not how we get configuration in these services.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
...
+    if (publicRoot == null || publicRoot === "") {
+      if (!hasWarnedMissingPublicStorageRoot) {
+        hasWarnedMissingPublicStorageRoot = true;
+        this.logger.warn(
```
roguenet: The module level `let` above to handle this case is not a great pattern. This is severe enough that instead of tracking this and doing a warning in the logs, go ahead and throw an `InternalServerErrorException`. This is a broken state and should be noticed asap.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
...
+          `Export file cleanup skipped: set ${PUBLIC_STORAGE_ROOT_ENV} when this service shares the filesystem used for generated exports.`
+        );
+      }
+      return;
+    }
+
+    const tempDir = path.join(publicRoot, "temp");
```
roguenet: We don't store files on the local file system in v3. Is that all the PHP code was doing? Exports get sent to S3, and I would have assumed that this was removing old exports from S3.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +185,115 @@ export class ScheduledJobsService {
...
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    await Task.sequelize!.transaction(async transaction => {
+      await this.runWithMysqlNamedLock(MYSQL_LOCK_RM_NOTIFICATIONS, transaction, async () => {
+        const cutoff = DateTime.utc().minus({ days: NOTIFICATION_RETENTION_DAYS }).toJSDate();
+        const removed = await Notification.destroy({
+          where: {
+            createdAt: { [Op.lte]: cutoff },
+            unread: false
+          },
+          transaction
```
roguenet: remove transaction.

Marcolr-dev:
Thanks for the review, addressed as follows: removed MySQL named locks and transactions for the verification, password-reset, and notification cleanup crons (safe to run concurrently); weekly digest crons still use named locks. Replaced sequelize! with the existing get sql() pattern on Task and ScheduledJob, and wired the service to use it. removeOldExportFiles no longer reads process.env or touches the local filesystem: it uses ConfigService for AWS_BUCKET, throws InternalServerErrorException if the bucket isn’t configured, and deletes stale objects under the temp/ prefix in S3 via a new FileService.deleteObjectsInPrefixOlderThan helper. Added unit tests for FileService; job-service tests still need a test DB in CI.

apps/job-service/src/scheduled-jobs/scheduled-jobs.service.ts
```diff
@@ -164,6 +178,55 @@ export class ScheduledJobsService {
...
+    const removed = await PasswordReset.destroy({
+      where: { createdAt: { [Op.lte]: cutoff } }
+    });
+    if (removed > 0) {
+      this.logger.log(`Removed ${removed} stale password resets (older than ${PASSWORD_RESET_RETENTION_DAYS}d)`);
+    }
+  }
+
+  @Cron(CronExpression.EVERY_DAY_AT_MIDNIGHT, { name: "removeOldExportFiles" })
+  async removeOldExportFiles(): Promise<void> {
```
roguenet: I went and reviewed the PHP code, and it is in fact just deleting temporary local files. We won't have those for exports in v3 - they contents will go straight to S3, so this one isn't needed after all. I double checked the bucket in S3, and we have every export ever created still there. Something we might want to address in the future but for now let's just remove this task.

roguenet (commented):
I'm surprised to see all these formatting changes. Why are they in here? Especially without a change to the linter config, I'd like to see that commit rolled back. The whole point of an opinionated linter is that it's the rule enforcer for if our code is readable and consistent and massive readability refactors like this should never be needed. If you'd like to advocate for a change to the linter config, we can talk about that, but a whole ton of pretty arbitrary changes is not something I'd like to see in an otherwise clean PR.

Marcolr-dev:
> I'm surprised to see all these formatting changes. Why are they in here? Especially without a change to the linter config, I'd like to see that commit rolled back. The whole point of an opinionated linter is that it's the rule enforcer for if our code is readable and consistent and massive readability refactors like this should never be needed. If you'd like to advocate for a change to the linter config, we can talk about that, but a whole ton of pretty arbitrary changes is not something I'd like to see in an otherwise clean PR.

The formatting commit was posted by mistake; I removed it from the branch history (reset + force push) and Origin is now aligned with the last good commit (1664758a). I apologize for the misunderstanding; I didn't notice the last commit that was pushed.

roguenet (approved):

PR #680, opened 2026-04-10
[TM-3107] add greenhouse notification

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-3107)

### DIFF
--- a/.env.local.sample
+++ b/.env.local.sample
@@ -53,3 +53,7 @@ APP_FRONT_END=http://localhost:3000
 # Should not be defined for local development.
 GA_API_SECRET=
 GA_MEASUREMENT_ID=
+
+# Greenhouse API credentials
+GREENHOUSE_API_URL=
+GREENHOUSE_API_TOKEN=
--- a/libs/common/src/lib/notifications/greenhouse-notification.service.ts
+++ b/libs/common/src/lib/notifications/greenhouse-notification.service.ts
@@ -1,0 +1,64 @@
+import { Injectable, InternalServerErrorException } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+import { Media, PolygonGeometry } from "@terramatch-microservices/database/entities";
+import { TMLogger } from "../util/tm-logger";
+
+@Injectable()
+export class GreenhouseNotificationService {
+  private readonly logger = new TMLogger(GreenhouseNotificationService.name);
+
+  private readonly apiUrl: string | null;
+  private readonly apiToken: string | null;
+
+  constructor(readonly configService: ConfigService) {
+    this.apiUrl = configService.get<string>("GREENHOUSE_API_URL") ?? null;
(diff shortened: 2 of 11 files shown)


libs/common/src/lib/notifications/greenhouse-notification.service.ts
```diff
@@ -0,0 +1,64 @@
...
+  async notifyMediaDeletion(media: Media) {
+    return this.runMutation("tmNotifyMediaDeleted", media.uuid);
+  }
+
+  async notifyPolygonUpdated(polygon: PolygonGeometry) {
+    return this.runMutation("tmNotifyPolygonUpdated", polygon.uuid);
+  }
+
+  private isEnabled() {
+    return (this.apiUrl != null || this.apiUrl != "") && this.apiToken != null && this.apiToken != "";
```
roguenet: These should all be `&&` (`this.apiUrl != null || this.apiUrl != ""` is always true)

libs/database/src/lib/hooks/media-destroy.ts
```diff
@@ -0,0 +1,30 @@
...
+      createdBy = mediaWithProperty?.getDataValue("createdBy") ?? null;
+    }
+    if (createdBy === null) {
+      return;
+    }
+    const user = await User.findByPk(createdBy, { include: [{ association: "roles" }] });
+    if (!user?.roles?.map(role => role.name).includes("greenhouse-service-account")) {
+      return;
+    }
+    await DatabaseModule.emitMediaDeleted(media);
```
roguenet, line 7: IMO, this hook should _always_ emit the event and the check for whether or not to contact Greenhouse should be handled in the event service. I could easily see adding code to the "media deleted" event handling and being very surprised that it only triggers if the media was owned by greenhouse.

.env.local.sample
```diff
@@ -53,3 +53,7 @@ APP_FRONT_END=http://localhost:3000
 # Should not be defined for local development.
 GA_API_SECRET=
 GA_MEASUREMENT_ID=
+
+# Greenhouse API credentials
+GREENHOUSE_API_URL=
+GREENHOUSE_API_TOKEN=
```
roguenet, line 59: The API token should be handled as a secret in github, which means you need to explicitly [copy it during deploy](https://github.com/wri/terramatch-microservices/blob/staging/.github/workflows/deploy-service.yml#L77-L87)

libs/common/src/lib/notifications/greenhouse-notification.service.ts
pachonjcl: changed

libs/database/src/lib/hooks/media-destroy.ts
pachonjcl: logic moved

.env.local.sample
pachonjcl: added that variable to the script

roguenet (approved):

PR #682, opened 2026-04-10
[TM-3255] Some DTOs are coming in with a numerical "UUID" instead of a string

roguenet:
https://gfw.atlassian.net/browse/TM-3255


roguenet:
Need to go ahead and merge this for the hotfix.

PR #685, opened 2026-04-13
[TM-3145] Refactor monetary input types to 'number-currency' for budg…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3145)


apps/entity-service/src/forms/forms.service.ts
```diff
@@ -177,14 +177,17 @@ export class FormsService {
...
       // For file questions, the collection is the collection of the field.
       const collection =
         (question.inputType === "file" ? (config?.field as LinkedFile | undefined)?.collection : question.collection) ??
         null;
       const childQuestions = sectionQuestions.filter(({ parentId }) => parentId === question.uuid);
       const options = optionsByQuestionId[question.id];
       const tableHeaders = tableHeadersByQuestionId[question.id];
-      return new FormQuestionDto(question, {
+      const questionForDto = Object.assign({}, question.get({ plain: true }), { inputType: resolvedInputType });
+      return new FormQuestionDto(questionForDto, {
```
roguenet: We have a pattern in place for omitting properties from the entity class so that they must be included in the second argument (the additional args). See: https://github.com/wri/terramatch-microservices/blob/staging/apps/entity-service/src/entities/dto/project.dto.ts#L122-L140

apps/entity-service/src/forms/forms.service.ts
```diff
@@ -177,14 +177,17 @@ export class FormsService {
 
     const questionToDto = (question: FormQuestion, sectionQuestions: FormQuestion[] = []) => {
       const config = getLinkedFieldConfig(question.linkedFieldKey ?? "");
+      const resolvedInputType =
+        question.linkedFieldKey != null && config?.field != null ? config.field.inputType : question.inputType;
```
roguenet: I'm not sure I see the point in this since you've got the migration in place to update the input type. We should be able to rely on `question.inputType` to be correct.

apps/entity-service/src/forms/forms.service.ts
roguenet: I think this won't be needed given the additional comment I just added.

libs/database/src/lib/migrations/202604131200-form-questions-monetary-input-type.ts
```diff
@@ -0,0 +1,24 @@
...
+    `);
+  },
+
+  async down({ context }) {
+    await context.sequelize.query(`
+      UPDATE form_questions
+      SET input_type = 'number'
+      WHERE linked_field_key IN ('pro-budget', 'pro-pit-bgt')
+        AND input_type = 'number-currency'
+    `);
```
roguenet: Ok, one more comment: this is just a data migration. Please do it in a one off instead. umzug should be used for schema migrations only.

roguenet (approved):

PR #701, opened 2026-04-22
[TM-3288] Implement monitoring partner projects retrieval in UsersSer…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3288)

### DIFF
--- a/libs/common/src/lib/dto/user.dto.ts
+++ b/libs/common/src/lib/dto/user.dto.ts
@@ -3,6 +3,7 @@ import { ApiProperty } from "@nestjs/swagger";
 import { JsonApiDto } from "../decorators";
 import { populateDto } from "./json-api-attributes";
 import { Framework, Project, User } from "@terramatch-microservices/database/entities";
+import { HybridSupportDto } from "./hybrid-support.dto";
 
 class UserFramework {
   @ApiProperty({ example: "TerraFund Landscapes" })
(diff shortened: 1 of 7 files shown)


LimberHope (dismissed):

LimberHope (dismissed):

libs/common/src/lib/dto/user.dto.ts
```diff
@@ -12,6 +12,22 @@ class UserFramework {
   slug: string;
 }
 
+/**
+ * Sideloaded `projects` resource on user show: monitoring-partner links (v2_project_users.is_monitoring).
+ */
+@JsonApiDto({ type: "projects" })
+export class ProjectMonitoringPartnerLightDto {
```
roguenet: `projects` is already in use, so this needs to be something else. However, I also think maybe this doesn't need to be a separate resource and can simply be embedded as an array of objects in the user response DTO. 

I also recommended on the FE PR that this start using the `lightResource` pattern, but I think maybe this should just get included in the index view as well so that all user objects sent to the client are complete.

Marcolr-dev: @roguenet Done: monitoring partners are embedded on UserDto as monitoringPartnerProjects (light DTOs with lightResource), no separate projects resource. Same field on list and show via getMonitoringPartnerProjectsByUserIds

roguenet (approved):

PR #713, opened 2026-04-29
[TM-3350] staging site report exports include links to live

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-3350)

### DIFF
--- a/apps/entity-service/src/entities/entities.service.ts
+++ b/apps/entity-service/src/entities/entities.service.ts
@@ -11,11 +11,7 @@ import {
   FormQuestion,
   Invasive,
   Media,
-  Nursery,
-  NurseryReport,
   Seeding,
-  Site,
-  SiteReport,
   Strata,
   Tracking,
   TreeSpecies,
@@ -29,7 +25,12 @@ import { EntityDto } from "./dto/entity.dto";
 import { AssociationProcessor } from "./processors/association-processor";
 import { AssociationDto, AssociationDtoAdditionalProps } from "@terramatch-microservices/common/dto/association.dto";
 import { NurseryProcessor } from "./processors/nursery.processor";
-import { ENTITY_MODELS, EntityModel, EntityType } from "@terramatch-microservices/database/constants/entities";
+import {
+  ENTITY_MODELS,
+  EntityModel,
+  EntityType,
+  isLinkedEntityModel
+} from "@terramatch-microservices/database/constants/entities";
 import { ProjectReportProcessor } from "./processors/project-report.processor";
 import { NurseryReportProcessor } from "./processors/nursery-report.processor";
 import { SiteReportProcessor } from "./processors/site-report.processor";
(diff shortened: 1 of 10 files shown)


apps/entity-service/src/entities/entities.service.ts
```diff
@@ -330,4 +338,12 @@ export class EntitiesService {
       }
     });
   }
+
+  private getTerramatchLink(type: EntityType, entity: Model, frontendUrl: string): string | undefined {
+    if (type === "sites") return (entity as Site).linkToTerramatch(frontendUrl);
+    if (type === "nurseries") return (entity as Nursery).linkToTerramatch(frontendUrl);
+    if (type === "siteReports") return (entity as SiteReport).linkToTerramatch(frontendUrl);
+    if (type === "nurseryReports") return (entity as NurseryReport).linkToTerramatch(frontendUrl);
+    return undefined;
```
roguenet: This is pretty awkward. In order to avoid mappings like this, you can first define a type (I'd put this in `entities.ts`, where a lot of similar things are held):
```
export type LinkedEntityModel = EntityModel & { linkToTerramatch: (frontEndUrl: string) => string };
export const isLinkedEntityModel = (model: EntityModel): model is LinkedEntityModel => "linkToTerramatch" in model;
```
Then instead of this mapping method, use this above:
```
          const additional = {
            ...(isLinkedEntityModel(entity) ? { linkToTerramatch: entity.linkToTerramatch(frontEndUrl) } : {}),
            ...(await this.csvExportService.collectFormCells(mappings, { [type]: entity }, frameworkKey)),
```

libs/database/src/lib/entities/nursery-report.entity.ts
```diff
@@ -127,8 +127,8 @@ export class NurseryReport extends Model<InferAttributes<NurseryReport>, InferCr
   @Column({ type: UUID, defaultValue: UUIDV4 })
   uuid: CreationOptional<string>;
 
-  get linkToTerramatch(): CreationOptional<string> {
-    return `https://www.terramatch.org/admin#/nurseryReport/${this.uuid}/show`;
+  linkToTerramatch(frontendUrl: string): CreationOptional<string> {
+    return `${frontendUrl}/admin#/nurseryReport/${this.uuid}/show`;
```
roguenet, line 131: `CreationOptional` is no longer needed now that this isn't a getter. I would leave the return type off entirely, since it's trivially determined by TS:
```
linkToTerramatch(frontendUrl: string) {
  return `${frontendUrl}/admin#/nurseryReport/${this.uuid}/show`;
}
```

apps/entity-service/src/entities/entities.service.ts
LimberHope: resolved

libs/database/src/lib/entities/nursery-report.entity.ts
LimberHope: resolved

roguenet (approved):

PR #714, opened 2026-04-30
[TM-3261] add passwordNotNull property for userDto

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-3261)

### DIFF
--- a/libs/common/src/lib/dto/user.dto.ts
+++ b/libs/common/src/lib/dto/user.dto.ts
@@ -45,7 +45,7 @@ export class UserDto {
         .filter(({ slug }) => slug != null)
         .map(({ name, slug }) => ({ name, slug })) as UserFramework[],
       monitoringPartnerProjects,
-      canLogIn: user.password == null
+      canLogIn: user.password != null
     });
   }
 
--- a/libs/common/src/lib/dto/user.dto.ts
+++ b/libs/common/src/lib/dto/user.dto.ts
@@ -45,7 +45,7 @@ export class UserDto {
         .filter(({ slug }) => slug != null)
         .map(({ name, slug }) => ({ name, slug })) as UserFramework[],
       monitoringPartnerProjects,
-      passwordNotNull: user.password != null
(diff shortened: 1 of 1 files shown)


roguenet (commented):
[Boolean naming](https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig) can be pretty important. Seeing a name like this makes me imagine some consumer writing `!passwordNotNull` which is pretty confusing. This would be much better as `passwordNull` or `passwordEmpty` or even `canLogIn` if you want to keep this value as true when it's not empty.

libs/common/src/lib/dto/user.dto.ts
```diff
@@ -45,7 +45,7 @@ export class UserDto {
         .filter(({ slug }) => slug != null)
         .map(({ name, slug }) => ({ name, slug })) as UserFramework[],
       monitoringPartnerProjects,
-      passwordNotNull: user.password != null
+      passwordEmpty: user.password != null
```
roguenet: This should be `== null`, or use `canLogIn: user.password != null`

LimberHope: using canLogIn

libs/common/src/lib/dto/user.dto.ts
```diff
@@ -45,7 +45,7 @@ export class UserDto {
         .filter(({ slug }) => slug != null)
         .map(({ name, slug }) => ({ name, slug })) as UserFramework[],
       monitoringPartnerProjects,
-      passwordEmpty: user.password != null
+      canLogIn: user.password == null
```
roguenet, line 48: This is still backwards - you made both changes! 😂 
```
canLogIn: user.password != null
```
The user can log in if their password has been set.

LimberHope: That’s strange
in the “Files changed” tab it looks as expected

roguenet (approved):

PR #731, opened 2026-05-06
[TM-3300] Enhance audit status handling with type filtering and new p…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3300)

### DIFF
--- a/apps/entity-service/src/entities/dto/audit-status-index-query.dto.ts
+++ b/apps/entity-service/src/entities/dto/audit-status-index-query.dto.ts
@@ -1,13 +1,25 @@
 import { ApiProperty } from "@nestjs/swagger";
-import { IsOptional, IsString } from "class-validator";
+import { Transform } from "class-transformer";
+import { IsArray, IsOptional, IsString } from "class-validator";
 
 export class AuditStatusIndexQueryDto {
   @ApiProperty({
     required: false,
+    isArray: true,
+    type: String,
     description:
-      "Comma-separated list of audit `type` values to include (e.g. polygon-data-submission,ready-for-baseline). When omitted, all audit rows are returned."
+      "Audit `type` values to include (e.g. repeat `types` query param). When omitted, all audit rows are returned."
   })
   @IsOptional()
-  @IsString()
-  types?: string;
+  @Transform(({ value }) => {
+    if (value == null || value === "") {
+      return undefined;
+    }
+    const raw = Array.isArray(value) ? value : [value];
+    const trimmed = raw.map((t: unknown) => String(t).trim()).filter(t => t.length > 0);
+    return trimmed.length > 0 ? trimmed : undefined;
+  })
+  @IsArray()
+  @IsString({ each: true })
+  types?: string[];
 }
--- a/apps/entity-service/src/entities/dto/audit-status-index-query.dto.ts
+++ b/apps/entity-service/src/entities/dto/audit-status-index-query.dto.ts
@@ -1,5 +1,4 @@
 import { ApiProperty } from "@nestjs/swagger";
-import { Transform } from "class-transformer";
 import { IsArray, IsOptional, IsString } from "class-validator";
 
 export class AuditStatusIndexQueryDto {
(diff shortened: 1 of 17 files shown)


apps/entity-service/src/entities/dto/audit-status-index-query.dto.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export class AuditStatusIndexQueryDto {
+  @ApiProperty({
+    required: false,
+    description:
+      "Comma-separated list of audit `type` values to include (e.g. polygon-data-submission,ready-for-baseline). When omitted, all audit rows are returned."
+  })
+  @IsOptional()
+  @IsString()
+  types?: string;
```
roguenet: Instead of using a comma separated list, the more typical pattern for this API is to use a `string[]` here. By the time it makes it to the controller, no further parsing should be required.

apps/unified-database-service/src/airtable/entities/project.airtable-entity.ts
```diff
@@ -125,7 +125,13 @@ const COLUMNS: ColumnMapping<Project, ProjectAssociations>[] = [
...
   "landownerAgreement",
   "bioeconomyProductList",
-  "bioeconomyProductDescription"
+  "bioeconomyProductDescription",
+  "polygonDataSubmission",
+  {
+    airtableColumn: "readyForBaseline",
+    dbColumn: "readyForBaseline",
+    valueMap: async ({ readyForBaseline }) => readyForBaseline === true
+  }
```
roguenet: Unless a ticket calls out for it explicitly, please don't add new columns to the airtable entity - this will break the sync.

libs/database/src/lib/migrations/index.ts
```diff
@@ -3,11 +3,11 @@ import { RunnableMigration } from "umzug";
...
 import { revertProjectBudgetColumnsProdTypes } from "./202604151400-revert-project-budget-columns-prod-types";
-import { makeAssortedColumnsNullable } from "./202605041703-make-assorted-columns-nullable";
+import { tm3300ProjectPolygonHandoff } from "./202605061200-tm-3300-project-polygon-handoff";
 
 export const migrations: RunnableMigration<QueryInterface>[] = [
   removeOrgConsortium,
   unifyMonetaryDecimal152,
   revertProjectBudgetColumnsProdTypes,
-  makeAssortedColumnsNullable
+  tm3300ProjectPolygonHandoff
```
roguenet, line 14: Please don't remove existing migrations.

apps/entity-service/src/entities/dto/audit-status-index-query.dto.ts
```diff
@@ -1,13 +1,25 @@
...
-  @IsString()
-  types?: string;
+  @Transform(({ value }) => {
+    if (value == null || value === "") {
+      return undefined;
+    }
+    const raw = Array.isArray(value) ? value : [value];
+    const trimmed = raw.map((t: unknown) => String(t).trim()).filter(t => t.length > 0);
+    return trimmed.length > 0 ? trimmed : undefined;
+  })
```
roguenet: Why is this needed? We have numerous examples of `string[]` query params that have been working for a long time (there are a bunch in the [site polygon query DTO](https://github.com/wri/terramatch-microservices/blob/3da0720074f0c01fbda6e20475cb17e9c865bd5e/apps/research-service/src/site-polygons/dto/site-polygon-query.dto.ts#L50-L58)) without a custom transform.

Marcolr-dev: You’re right :this isn’t needed for consistency with our other string[] query DTOs. The web client serializes array params with qs arrayFormat: "indices", so types arrives as a real array. The transform was only handling the edge case of a bare types=single string (and trim/empty), which we don’t rely on elsewhere. Removed the @Transform to match EntityQueryDto / TrackingsQueryDto style.

LimberHope (approved):

apps/unified-database-service/src/airtable/entities/project.airtable-entity.ts
roguenet: Ah, I missed that the ticket did indeed call out the addition to airtable. That's a new pattern - we usually add airtable columns in a separate ticket.

PR #735, opened 2026-05-08
[TM-3371] test hook update

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-3371)

### DIFF
--- a/libs/common/src/lib/linkedFields/linkedAnswerCollector/field.collector.ts
+++ b/libs/common/src/lib/linkedFields/linkedAnswerCollector/field.collector.ts
@@ -241,7 +241,17 @@ export function fieldCollector(logger: LoggerService): FieldResourceCollector {
         } else {
           // make sure it hasn't been used for a typical demographics entry, as in that case we
           // don't want to handle trying to balance with this single integer value.
-          const entries = await TrackingEntry.tracking(tracking.id).findAll();
+          let entries = await TrackingEntry.tracking(tracking.id).findAll();
+          const hasComplicatedEntries =
+            entries.length !== virtual.entryTypes.length ||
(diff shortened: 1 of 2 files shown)


libs/common/src/lib/linkedFields/linkedAnswerCollector/field.collector.ts
```diff
@@ -241,7 +241,17 @@ export function fieldCollector(logger: LoggerService): FieldResourceCollector {
...
+          const hasComplicatedEntries =
+            entries.length !== virtual.entryTypes.length ||
+            entries.find(({ subtype }) => subtype !== "unknown") != null;
+          if (hasComplicatedEntries && virtual.domain === "demographics" && question.inputType === "number") {
+            await TrackingEntry.destroy({ where: { id: entries.map(({ id }) => id) } });
+            await TrackingEntry.bulkCreate(
+              virtual.entryTypes.map(type => ({ trackingId: tracking.id, type, subtype: "unknown", amount: value }))
+            );
+            entries = await TrackingEntry.tracking(tracking.id).findAll();
+          }
```
roguenet, line 254: In case we end up going this route (see conversation with Ben in slack), I'll go ahead and leave some feedback.
* I wouldn't isolate this fix to just demographics, and checking the input type is not necessary. This change would constitute a strategy shift in how we deal with this situation and it should apply universally. It's worth noting that those two conditions (demographics and number input) are true for all uses of this virtual field type right now anyway.
* No need to do the `findAll`, just keep the result from `bulkCreate`. Or actually, after removing the block below (next bullet) you won't need a reference to `entries` at all.
* You'd remove the block below that checks for complicated entries and throws an error; this code replaces that functionality.

PR #738, opened 2026-05-11
[TM-3338] Add Overlap Filter

cesarLima1:
https://gfw.atlassian.net/browse/TM-3338


apps/research-service/src/site-polygons/site-polygon-query.builder.ts
```diff
@@ -227,6 +228,24 @@ export class SitePolygonQueryBuilder extends PaginatedQueryBuilder<SitePolygon>
     return this;
   }
 
+  filterHasOverlap(hasOverlap?: boolean) {
+    if (hasOverlap !== true) return this;
+    const sequelize = SitePolygon.sequelize;
+    if (sequelize == null) {
+      throw new BadRequestException("Database connection not available");
+    }
```
roguenet: Please use the `static get sql()` pattern we've [established in many other records](https://github.com/wri/terramatch-microservices/blob/staging/libs/database/src/lib/entities/point-geometry.entity.ts#L9)

cesarLima1: Done, added `static get sql` on `SitePolygon`

roguenet (approved):

PR #766, opened 2026-05-26
[TM-3404] Bulk Edit Polygons

cesarLima1:
https://gfw.atlassian.net/browse/TM-3404

### DIFF
--- a/apps/research-service/src/site-polygons/site-polygon-creation.service.ts
+++ b/apps/research-service/src/site-polygons/site-polygon-creation.service.ts
@@ -14,7 +14,7 @@ import {
   AttributeChangesDto,
   CreateSitePolygonRequestDto
 } from "./dto/create-site-polygon-request.dto";
-import { SitePolygonBulkAttributeChangesDto } from "./dto/site-polygon-bulk-attribute-update.dto";
+import type { SitePolygonBulkAttributeChangesDto } from "./dto/site-polygon-bulk-attribute-update.dto";
 
 const BULK_ATTRIBUTE_UPDATE_CHANGE_REASON = "Bulk attribute update via API" as const;
 import { PolygonGeometryCreationService } from "./polygon-geometry-creation.service";
@@ -30,6 +30,7 @@ import {
 import { DuplicateGeometryValidator } from "../validations/validators/duplicate-geometry.validator";
 import {
   CriteriaId,
+  SITE_POLYGON_TARGET_SYSTEMS,
   VALIDATION_CRITERIA_IDS,
   CRITERIA_ID_TO_VALIDATION_TYPE,
   ValidationType
@@ -42,6 +43,66 @@ import "multer";
 interface DuplicateCheckResult {
   duplicateIndexToUuid: Map<number, string>;
 }
+
+const validateAndSortStrictStringArray = (
+  value: unknown,
+  validValues: readonly string[],
+  fieldName: string
+): string[] | null => {
+  if (value == null) {
+    throw new BadRequestException(`${fieldName} cannot be null`);
+  }
+  if (!Array.isArray(value)) {
+    throw new BadRequestException(`${fieldName} must be an array`);
+  }
+  if (value.length === 0) return null;
+
+  const nonStringValues = value.filter(item => typeof item !== "string");
+  if (nonStringValues.length > 0) {
+    throw new BadRequestException(`${fieldName} must contain only strings`);
+  }
+
+  const trimmedValues = value.map(item => item.trim()).filter(item => item !== "");
+  if (trimmedValues.length === 0) return null;
+
+  const validValuesSet = new Set(validValues);
+  const invalidValues = trimmedValues.filter(item => !validValuesSet.has(item));
+  if (invalidValues.length > 0) {
+    throw new BadRequestException(`${fieldName} contains invalid value(s): ${invalidValues.join(", ")}`);
+  }
+
+  return validateAndSortStringArray(trimmedValues, validValues);
+};
+
+const validateTargetSystem = (value: unknown): string | null => {
+  if (value == null) {
+    throw new BadRequestException("targetSys cannot be null");
+  }
+  if (typeof value !== "string") {
+    throw new BadRequestException("targetSys must be a string");
+  }
+  const trimmedValue = value.trim();
+  if (trimmedValue === "") return null;
+  if (!SITE_POLYGON_TARGET_SYSTEMS.includes(trimmedValue as (typeof SITE_POLYGON_TARGET_SYSTEMS)[number])) {
+    throw new BadRequestException(`targetSys contains invalid value: ${trimmedValue}`);
+  }
+  return trimmedValue;
+};
+
+const validatePlantStart = (value: unknown): Date | null => {
+  if (value == null) {
+    throw new BadRequestException("plantStart cannot be null");
+  }
+  if (typeof value !== "string") {
+    throw new BadRequestException("plantStart must be a string");
+  }
+  if (value.trim() === "") return null;
+  const date = new Date(value);
+  if (Number.isNaN(date.getTime())) {
+    throw new BadRequestException(`plantStart contains invalid date: ${value}`);
+  }
+  return date;
+};
 
 interface ValidationIncludedData {
   type: "validation";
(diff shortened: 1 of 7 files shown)


apps/research-service/src/site-polygons/site-polygon-creation.service.ts
```diff
@@ -40,6 +44,66 @@ interface DuplicateCheckResult {
...
 }
 
+const validateAndSortStrictStringArray = (
+  value: unknown,
+  validValues: readonly string[],
+  fieldName: string
+): string[] | null => {
+  if (value == null) {
+    throw new BadRequestException(`${fieldName} cannot be null`);
+  }
```
roguenet: This case is covered by the array check below and can be skipped.

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
```diff
@@ -40,6 +44,66 @@ interface DuplicateCheckResult {
...
+  if (!Array.isArray(value)) {
+    throw new BadRequestException(`${fieldName} must be an array`);
+  }
+  if (value.length === 0) return null;
+
+  const nonStringValues = value.filter(item => typeof item !== "string");
+  if (nonStringValues.length > 0) {
```
roguenet: This would be a little more efficient with find instead of filter because it'll stop on the first non string value:
```
const hasNonString = value.find(item => typeof item !== "string") != null;
if (hasNonString) {
```

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
```diff
@@ -40,6 +44,66 @@ interface DuplicateCheckResult {
...
+    throw new BadRequestException(`${fieldName} contains invalid value(s): ${invalidValues.join(", ")}`);
+  }
+
+  return validateAndSortStringArray(trimmedValues, validValues);
+};
+
+const validateTargetSystem = (value: unknown): string | null => {
+  if (value == null) {
+    throw new BadRequestException("targetSys cannot be null");
+  }
```
roguenet: Same here, I don't think the null check is necessary

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
```diff
@@ -40,6 +44,66 @@ interface DuplicateCheckResult {
...
+  if (!SITE_POLYGON_TARGET_SYSTEMS.includes(trimmedValue as (typeof SITE_POLYGON_TARGET_SYSTEMS)[number])) {
+    throw new BadRequestException(`targetSys contains invalid value: ${trimmedValue}`);
+  }
+  return trimmedValue;
+};
+
+const validatePlantStart = (value: unknown): Date | null => {
+  if (value == null) {
+    throw new BadRequestException("plantStart cannot be null");
+  }
```
roguenet: Skip null check

apps/research-service/src/site-polygons/site-polygons.controller.spec.ts
```diff
@@ -733,7 +791,222 @@ describe("SitePolygonsController", () => {
...
+
+      const originalSequelize = SitePolygon.sequelize;
+      const mockSequelize = {
+        transaction: jest.fn((callback: (transaction: object) => unknown) => callback({}))
+      };
+      Object.defineProperty(SitePolygon, "sequelize", {
+        value: mockSequelize,
+        writable: true,
+        configurable: true
+      });
```
roguenet: Why is this needed? Specs should interact with the database, including transactions...

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -402,6 +404,112 @@ export class SitePolygonsController {
...
+  })
+  @JsonApiResponse({ data: SitePolygonLightDto, hasMany: true })
+  @ExceptionResponse(UnauthorizedException, { description: "Authentication failed." })
+  @ExceptionResponse(BadRequestException, { description: "Invalid request data or empty attributeChanges." })
+  @ExceptionResponse(NotFoundException, { description: "At least one of the site polygons was not found." })
+  async bulkUpdateAttributes(@Body() request: SitePolygonBulkAttributeUpdateBodyDto) {
+    const { data, attributeChanges } = request;
+
+    if (attributeChanges == null || Object.keys(attributeChanges).length === 0) {
+      throw new BadRequestException("attributeChanges must contain at least one field to update");
```
roguenet: This should be handled by the DTO validation decorators.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -402,6 +404,112 @@ export class SitePolygonsController {
...
+    }
+
+    const unsupportedKeys = Object.keys(attributeChanges).filter(
+      key => !BULK_ATTRIBUTE_CHANGE_KEYS.includes(key as (typeof BULK_ATTRIBUTE_CHANGE_KEYS)[number])
+    );
+    if (unsupportedKeys.length > 0) {
+      throw new BadRequestException(`Unsupported attributeChanges field(s): ${unsupportedKeys.join(", ")}`);
```
roguenet: Same here; this should be handled in the DTO

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -402,6 +404,112 @@ export class SitePolygonsController {
...
+    }
+
+    for (const sitePolygon of sitePolygons) {
+      await this.policyService.authorize("update", sitePolygon);
+    }
+
+    const userId = this.policyService.userId;
+    if (userId == null) {
+      throw new UnauthorizedException("User must be authenticated");
+    }
```
roguenet: This is not possible on endpoints that require auth. Just use `userId as number` below to deal with the TS error.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -402,6 +404,112 @@ export class SitePolygonsController {
...
+    const user = await User.findByPk(userId, {
+      attributes: ["firstName", "lastName"],
+      include: [{ association: "roles", attributes: ["name"] }]
+    });
+    const source = user?.getSourceFromRoles() ?? "terramatch";
+    const userFullName = user?.fullName ?? null;
+
+    if (SitePolygon.sequelize == null) {
+      throw new BadRequestException("Database connection not available");
+    }
```
roguenet: This is not necessary. Just use `await SitePolygon.sql.transaction` below; that getter handles throwing if the connection is not there. (and it uses a more appropriate error - this is not a bad request situation, it's an internal error)

cesarLima1: Removed, now using `SitePolygon.sql.transaction`.

apps/research-service/src/site-polygons/site-polygons.controller.ts
cesarLima1: Removed, using `this.policyService.userId` as number since auth is guaranteed on this endpoint.

apps/research-service/src/site-polygons/site-polygons.controller.ts
cesarLima1: Moved to DTO.

apps/research-service/src/site-polygons/site-polygons.controller.ts
cesarLima1: Moved to DTO.

apps/research-service/src/site-polygons/site-polygons.controller.spec.ts
cesarLima1: Removed the mock, tests now use the real `SitePolygon.sql.transaction` from the test DB.

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
cesarLima1: Done, removed the null check.

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
cesarLima1: Done, removed the null check.

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
cesarLima1: Done — switched to `find`.

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
cesarLima1: Done, removed the redundant null check.

apps/research-service/src/site-polygons/site-polygons.controller.ts
```diff
@@ -452,10 +440,7 @@ export class SitePolygonsController {
       await this.policyService.authorize("update", sitePolygon);
     }
 
-    const userId = this.policyService.userId;
-    if (userId == null) {
-      throw new UnauthorizedException("User must be authenticated");
-    }
+    const userId = this.policyService.userId as number;
```
roguenet, line 443: :art: Sorry I didn't catch this earlier (I won't require you change it because this is effectively the same); code doesn't need to rely on any service to get the current in-context user. `UserContext.authenticatedUserId` is the correct way to do this now: https://github.com/wri/terramatch-microservices/blob/staging/libs/common/src/lib/contexts/user.context.ts#L16

roguenet (approved):

apps/research-service/src/site-polygons/site-polygons.controller.ts
cesarLima1: I’ll keep that in mind for future PRs.

PR #771, opened 2026-05-29
[TM-3355] Add import for project cohort landscape migration

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3355)

### DIFF
--- a/apps/entity-service/src/repl/oneOff/2026-05-28-importProjectCohortLandscape.ts
+++ b/apps/entity-service/src/repl/oneOff/2026-05-28-importProjectCohortLandscape.ts
@@ -1,0 +1,159 @@
+import { columnValue, parseCsv } from "@terramatch-microservices/common/util/repl/csv";
+import { withoutSqlLogs } from "@terramatch-microservices/common/util/repl/without-sql-logs";
+import { Project } from "@terramatch-microservices/database/entities";
+import { isEqual } from "lodash";
+
+type ProjectCohortLandscapeRow = {
+  uuid: string;
+  cohort: string;
+  landscape: string | null;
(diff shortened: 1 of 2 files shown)


apps/entity-service/src/repl/oneOff/2026-05-28-importProjectCohortLandscape.ts
```diff
@@ -0,0 +1,159 @@
...
+    return TM_3355_PROJECT_COHORT_LANDSCAPE_DATA;
+  }
+
+  const rows: ProjectCohortLandscapeRow[] = [];
+  await parseCsv(csvPath, async row => {
+    rows.push({
+      uuid: columnValue(row, "uuid") as string,
+      cohort: columnValue(row, "cohort") as string,
+      landscape: columnValue(row, "landscape")
+    });
```
roguenet, line 85: I imagine this CSV isn't huge so no need to change for this one, but in general it's a good idea to remember that the AWS machines are low-resource, meaning they don't have a ton of memory. That's why I usually process each row at a time instead of stuffing the whole CSV into an array.

roguenet (approved):

apps/entity-service/src/repl/oneOff/2026-05-28-importProjectCohortLandscape.ts
Marcolr-dev: Okay, for the TM-3355 there are 17 fixed registers, so the array in memory isn't a problem. In future imports with large CSV files, I'll follow the pattern of processing row by row within `parseCsv` without accumulating.

PR #793, opened 2026-06-08
[TM-3517] text translation for exports

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-3517)


libs/common/src/lib/localization/localization.service.ts
```diff
@@ -143,8 +144,20 @@ export class LocalizationService {
...
+    if (!tx.cache.hasTranslations(txLocale)) {
+      await tx.fetchTranslations(locale);
+    }
+
+    const txTranslations = tx.cache.getTranslations(locale);
+    const hash = md5(text);
+    const translation = txTranslations[hash];
+    if (translation != null) {
+      return translation;
+    }
```
roguenet, line 162: Left a comment in Slack, but wanted to mention here as well: This code path ignores `params`, and I'm worried it's going to break some existing usage of this method. This was a nice to have for the release, so let's not merge into the release at this time, and try to figure out a better solution for GG.

roguenet (approved):

PR #799, opened 2026-06-11
[TM-3469] implement server side ga4 for polygon creation

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-3469)

### DIFF
--- a/libs/common/src/lib/events/event.service.ts
+++ b/libs/common/src/lib/events/event.service.ts
@@ -64,7 +64,7 @@ export class EventService {
   }
 
   async sendPolygonPushedViaApiAnalytics(partnerId: string, params: PolygonPushedViaApiParams) {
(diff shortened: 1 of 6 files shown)


libs/common/src/lib/events/event.service.ts
```diff
@@ -61,4 +62,14 @@ export class EventService {
       params: { modelType: modelLaravelType, status }
     });
   }
+
+  async sendPolygonPushedViaApiAnalytics(partnerId: string, params: PolygonPushedViaApiParams) {
+    this.logger.log(
```
pachonjcl: I would prefer this to be debug not log level

pachonjcl (changes requested):

pachonjcl (approved):

PR #800, opened 2026-06-12
[TM-3523] Add embeddedDocumentationDto method to MediaService and upd…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3523)

### DIFF
--- a/libs/common/src/lib/media/media.service.ts
+++ b/libs/common/src/lib/media/media.service.ts
@@ -163,7 +163,7 @@ export class MediaService {
   public getUrl(media: Media, conversion?: string) {
     const endpoint = this.endpoint;
     if (conversion == null) return `${endpoint}${this.filePath(media)}`;
-    return media.generatedConversions[conversion] == null
+    return media.generatedConversions?.[conversion] == null
       ? null
       : `${endpoint}/${this.conversionFilePath(media, conversion)}`;
   }
--- a/libs/common/src/lib/media/media.service.ts
+++ b/libs/common/src/lib/media/media.service.ts
@@ -373,13 +373,9 @@ export class MediaService {
     return media;
   }
 
-  getPresignedDownloadUrl(media: Media) {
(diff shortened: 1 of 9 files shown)


libs/common/src/lib/media/media.service.ts
```diff
@@ -372,6 +373,17 @@ export class MediaService {
     return media;
   }
 
+  getPresignedDownloadUrl(media: Media) {
```
roguenet: Is this method being used by any other callers? If not, please just inline to the dto method below. No reason to add it this public interface of this service unless there's an external need for it.

libs/common/src/lib/media/media.service.ts
```diff
@@ -372,6 +373,17 @@ export class MediaService {
...
+    return this.fileService.generatePresignedUrl(this.bucket, `${media.id}/${media.fileName}`);
+  }
+
+  async embeddedDocumentationDto(media: Media) {
```
roguenet: Why is this called `embeddedDocumentationDto`? It would work for any media, the difference is that it uses a presigned URL instead of a direct S3 URL. So maybe something more like `presignedMediaDto`

roguenet (commented):
I don't understand the need for the presigned URL in this PR. I sure thought that any documentation that's being stored from users is going to the bucket that doesn't need presigned URLs to download from.

libs/common/src/lib/media/media.service.ts
Marcolr-dev: You're right , user documentation lives in the public bucket (ACL: public-read), same as other form file uploads via file.collector. Presigned URLs weren't needed here. I've removed getPresignedDownloadUrl, switched to direct S3 URLs via getUrl(), and renamed the method to embeddedMediaDto since it's not documentation-specific.

roguenet (approved):

PR #817, opened 2026-06-22
Feat/[TM-3592] implement new header style to all reports pages

LimberHope:
[Task Link](https://gfw.atlassian.net/browse/TM-3592)


apps/entity-service/src/entities/processors/srp-report.processor.ts
```diff
@@ -209,4 +212,68 @@ export class SrpReportProcessor extends ReportProcessor<
...
+          endMonth: endMonthName,
+          year
+        });
+      } else {
+        return await getRangeTitle();
+      }
+    } else {
+      return await getRangeTitle();
+    }
+  }
```
pachonjcl: instead of nesting, do a early return

apps/entity-service/src/entities/processors/financial-report.processor.ts
```diff
@@ -227,4 +230,68 @@ export class FinancialReportProcessor extends ReportProcessor<
...
+          endMonth: endMonthName,
+          year
+        });
+      } else {
+        return await getRangeTitle();
+      }
+    } else {
+      return await getRangeTitle();
+    }
+  }
```
pachonjcl: the piece of code below seems to be identical or very similar, please make a reusable function

pachonjcl (changes requested):

pachonjcl (approved):

PR #848, opened 2026-07-10
[TM-3426] deleted polygons

egrojMonroy:
https://gfw.atlassian.net/browse/TM-3426


apps/research-service/src/site-polygons/deleted-site-polygon-query.builder.ts
```diff
@@ -0,0 +1,54 @@
+import { Op, WhereOptions } from "sequelize";
+import { PaginatedQueryBuilder } from "@terramatch-microservices/common/util/paginated-query.builder";
+import { Disturbance, PolygonGeometry, Site, SitePolygon } from "@terramatch-microservices/database/entities";
+import { SITE_POLYGON_SEARCH_FIELDS, SitePolygonSearchField } from "./dto/site-polygon-query.dto";
+
+export class DeletedSitePolygonQueryBuilder extends PaginatedQueryBuilder<SitePolygon> {
```
roguenet: I'm not sure I see a need for this subclass. It doesn't appear to be adding any functionality - the logic for constructing things like the search where terms can be handled in the service where this is being used.

The ability to set `paranoid: false` on the find options can simply be added as a feature to the base query builder.

egrojMonroy: changed and deleted

cesarLima1 (approved):

PR #852, opened 2026-07-14
[TM-3295] submission cycle single

egrojMonroy:
https://gfw.atlassian.net/browse/TM-3295?focusedCommentId=85837

### DIFF
--- a/apps/research-service/src/site-polygons/site-polygon-creation.service.ts
+++ b/apps/research-service/src/site-polygons/site-polygon-creation.service.ts
@@ -86,6 +86,18 @@ const validateTargetSystem = (value: unknown): string | null => {
   return trimmedValue;
 };
 
+const validateSubmissionCycle = (value: unknown): string | null => {
+  if (typeof value !== "string") {
+    throw new BadRequestException("submissionCycle must be a string");
+  }
+  const trimmedValue = value.trim();
+  if (trimmedValue === "") return null;
+  if (!SITE_POLYGON_SUBMISSION_CYCLES.includes(trimmedValue as (typeof SITE_POLYGON_SUBMISSION_CYCLES)[number])) {
+    throw new BadRequestException(`submissionCycle contains invalid value: ${trimmedValue}`);
+  }
+  return trimmedValue;
+};
+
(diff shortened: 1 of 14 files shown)


apps/research-service/src/site-polygons/utils/site-polygon-property-validator.ts
```diff
@@ -175,9 +172,8 @@ export function validateAndSortStringArray(
   return filteredValues.length > 0 ? filteredValues.sort() : null;
 }
 
-function validateTargetSys(value: string): string | null {
+function trimToNullableString(value: string): string | null {
   if (value == null || value.trim().length === 0) return null;
 
-  const trimmed = value.trim();
-  return trimmed.length > 0 ? trimmed : null;
+  return value.trim();
```
roguenet: 🎨 Totally not necessary to change, but when I see a function getting executed twice like this, it gives me an itch. I would do it like this:
```
const result = (value ?? "").trim():
return result.length === 0 ? null : result;
```

apps/research-service/src/site-polygons/site-polygon-creation.service.ts
```diff
@@ -86,6 +86,18 @@ const validateTargetSystem = (value: unknown): string | null => {
   return trimmedValue;
 };
 
+const validateSubmissionCycle = (value: unknown): string | null => {
+  if (typeof value !== "string") {
+    throw new BadRequestException("submissionCycle must be a string");
+  }
```
roguenet: Since this `value` appears to be coming from validated DTOs, is there any reason this check is needed? Seems like the method should be able to accept `value: string` or maybe `value: string | null | undefined`... I'd have to dig more to be sure.

egrojMonroy: trust the DTO, removed this

apps/research-service/src/site-polygons/utils/site-polygon-property-validator.ts
egrojMonroy: removed

pachonjcl (approved):

PR #859, opened 2026-07-15
[TM-3637] add qa polygon columns on airtable

pachonjcl:
[Task Link](https://gfw.atlassian.net/browse/TM-3637)


roguenet (approved):
I'll get the columns set up in Airtable. Once this deploys, ping me on slack to test in the staging AT environment.

PR #862, opened 2026-07-16
[TM-3677] Add bioeconomy fields to project and project report entitie…

Marcolr-dev:
[Task Link](https://gfw.atlassian.net/browse/TM-3677)


apps/entity-service/src/repl/oneOff/2026-07-16-fundoFloraProjectReportBioeconomyQuestions.ts
```diff
@@ -0,0 +1,270 @@
...
+    created = true;
+  }
+
+  const optionsCreated =
+    seed.options == null ? 0 : await seedQuestionOptions(localizationService, question, seed.options);
+
+  return { created, optionsCreated };
+}
+
+export const fundoFloraProjectReportBioeconomyQuestions = withoutSqlLogs(
```
roguenet: A couple of issues here: 
* This should be creating new FormOptionLists to match the new `optionListKey`s in the configs (like `bioeconomy-product-list`)
* This should not be creating new form sections or questions - that is handled in the form editor once the system additions are in place.

When a form question is added to a section in the admin editor, it uses the FormOptionList to populate the default values.

apps/entity-service/src/entities/form-data.service.ts
```diff
@@ -244,6 +249,20 @@ export class FormDataService {
     return document;
   }
 
+  async buildFormModels(modelType: FormModelType, entity: EntityModel): Promise<FormModels> {
```
roguenet: Would you please explain the need for this? A new pattern like this is surprising to me for this ticket.

Marcolr-dev: buildFormModels was added because the ticket placed women_governance on v2_projects while using a pro-rep-* linked key on the project report form — that required loading the parent project when saving report answers. We've removed it and store women_governance on v2_project_reports instead, matching the other pro-rep-* fields. If the ticket strictly requires v2_projects, we can discuss restoring cross-entity sync

roguenet: Really, a project report should never be saving data back to the project, so if that ends up being something asked of you, please loop me into the discussion.

roguenet (approved):

